PR #4, opened 2020-02-10
compilation fixes

hannesm:
best viewed commit by commit. this is supposed to achieve the following (by cherry-picking and polishing commits from a multitude of forks):
- use dune instead of ocamlbuild/topkg
- revert the "remove sexplib serialisers"
- use `dune-configurator` and `bigarray-compat`
- do not use `-Wno-implicit-fallthrough`, since this is a reasonable warning. instead, mark the case statements explicitly where a fallthrough is expected (use a C comment instead of annotation to make gcc<7 and clang not complain)
- fix "unused C function" warnings
- cross-compilation for freestanding and xen (by flattening the `src/native` directory and copying around)
- be more precise why `warning -3` is in nocrypto_lwt_entropy

the freestanding and xen support ~~is not yet properly tested (and may need some more work)~~ works fine (a static_website_tls unikernel links, and the freestanding/hvt one also runs smoothly), I'll follow up on that. once this is good to go, the renaming could happen.


dune-project
```diff
@@ -0,0 +1,3 @@
+(lang dune 1.7)
+(name nocrypto)
+(version %%VERSION_NUM%%)
```
emillon: this is not necessary, `dune subst` will populate that when pinning and `dune-release` will insert it at release time.

config/cfg.ml
```diff
@@ -0,0 +1,12 @@
...
+let flags = ["-DACCELERATE"; "-mssse3"; "-maes"; "-mpclmul"]
+
+let _ =
+  let auto = match Cpuid.supports needs with Ok true -> flags | _ -> [] in
+  let fs = match Sys.getenv evar with
+    "true" -> flags
+  | "false" -> []
+  | _ -> auto
+  | exception Not_found -> auto in
+  Format.(printf "(@[%a@])%!" (fun ppf -> List.iter (fprintf ppf "%s@ ")) fs)
```
emillon: You can use Configurator.V1.Flags.write_sexp here.

hannesm: ok, I do not know how to do this. I am unable to find the documentation / API of `Configurator.V1.Flags`. this `cfg.exe` is used by `src/dune` (with a rule `(rule (with-stdout-to c_flags.sexp (run ../config/cfg.exe)))`) as well as xen/dune and freestanding/dune.

From https://dune.readthedocs.io/en/stable/quick-start.html#defining-a-library-with-c-stubs I gather that your proposed `write_sexp` takes two arguments, a file name and an OCaml value - but it is rather unclear to me where the output is written to, then (in terms of relative path).

.gitignore
```diff
@@ -2,11 +2,15 @@ _build
 *.install
```
avsm: can remove the  (native, byte) from here as dune places those in _build

.ocamlinit
```diff
@@ -1,13 +1,14 @@
...
-#directory "_build/src"
-#load "nocrypto.cma"
-
-(* #require "oUnit" *)
-(* #directory "_build/tests" *)
-(* #load "testlib.cma" *)
+#directory "_build/default/src"
+#directory "_build/default/src/.nocrypto.objs/byte"
+#directory "_build/default/unix"
+#directory "_build/default/unix/.nocrypto_entropy_unix.objs/byte"
```
avsm: This file shouldn't be needed now either, as `dune utop` should do the trick to generate the toplevel

.travis.yml
```diff
@@ -7,7 +7,6 @@ env:
     - UPDATE_GCC_BINUTILS=1
     - TESTS=true
   matrix:
-    - OCAML_VERSION=4.02
     - OCAML_VERSION=4.03
     - OCAML_VERSION=4.04
     - OCAML_VERSION=4.05
```
avsm: We can remove everything <4.06 from here I believe...

lwt/nocrypto_entropy_lwt.ml
```diff
@@ -1,19 +1,23 @@
-open Lwt
+open Lwt.Infix
 open Nocrypto
 
+[@@@ocaml.warning "-3"]
+(* Used for hooks, https://github.com/ocsigen/lwt/issues/573 *)
+module Lwt_sequence = Lwt_sequence
```
avsm: This can be Lwt_dllist as well, but that can happen in a subsequent PR

nocrypto.opam
```diff
@@ -0,0 +1,36 @@
...
+license:      "ISC"
+synopsis:     "Simple crypto for the modern age"
+
+build: [ ["dune" "subst"] {pinned}
+         ["dune" "build" "-p" name "-j" jobs ]
+         ["dune" "runtest"] {with-test} ]
+
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build & >= "1.7"}
```
avsm:
```suggestion
  "dune" {>= "1.7"}
```

.gitignore
hannesm: thanks, done

dune-project
hannesm: thanks, done

.ocamlinit
hannesm: ok, done

.travis.yml
hannesm: I actually removed travis.yml, and we'll focus on ocaml-ci here :)

nocrypto.opam
hannesm: already done :)

hannesm:
thanks for your reviews.

I force-pushed to add a lower bound of OCaml 4.07.0 -- and then removed the `bigarray-compat` commit (since this is only needed < 4.07.0 -- in 4.07.0+ we have Stdlib.Bigarray)

lwt/nocrypto_entropy_lwt.ml
hannesm: thanks; please take a look into 45e7a86 for an alternative solution -- no deprecated API, no dllist

hannesm:
hmm, now when I run `dune utop`, I get:
```
File "xen/dune", line 7, characters 12-28:
7 |  (libraries mirage-xen-posix)
                ^^^^^^^^^^^^^^^^
Error: Library "mirage-xen-posix" not found.
Hint: try: dune external-lib-deps --missing .utop/utop.exe
```

but in `xen/dune` the stanza `optional` is included -- any dune hints how to get `dune utop` to work in such a scenario (dune 2.1.3 is what I use)?

another question about CI: is it possible that it runs with optional dependencies? (in case the answer is no (what I expect), I guess an appropriate solution is to have separate packages (though I'm not sure how this would work with the `META.in`, but I can try to figure that out).

talex5:
> another question about CI: is it possible that it runs with optional dependencies? (in case the answer is no (what I expect), I guess an appropriate solution is to have separate packages (though I'm not sure how this would work with the META.in, but I can try to figure that out).

Yeah, separate packages would be great!

If you need to force optional deps to be installed in CI as a quick fix though you could always add a dummy opam package that depends on all the optional stuff. Or you could make the optional dependencies required for tests.

hannesm:
@talex5 thanks for your quick and valuable response. Since I'm not eager to accumulate hacks in this repository, but will subsequently revise the package layout (not in this PR).

PR #5, opened 2020-02-14
split into three packages (mirage-crypto, mirage-crypto-rng, mirage-crypto-pk)

hannesm:
~~still work in progress~~, this is a split into three opam packages. ~~need to get the tests running, and a sweep through the source.~~


mirage-crypto-pk.opam
```diff
@@ -0,0 +1,29 @@
...
+license:      "ISC"
+synopsis:     "Simple public-key cryptography for the modern age"
+
+build: [ ["dune" "subst"] {pinned}
+         ["dune" "build" "-p" name "-j" jobs ]
+         ["dune" "runtest" "-p" name "-j" jobs] {with-test} ]
+
+depends: [
+  "ocaml" {>= "4.07.0"}
+  "dune" {>= "1.7"}
```
dinosaure: should we use `dune.2.0.0`?

hannesm:
I added a macos travis runner (fingers crossed that has macos >= 10.12)

hannesm:
this is now ready for review -- I squashed most of the commits into a single one (see commit message for details).

in addition, this PR adds CI runners for macOS (via travis) and FreeBSD (via cirrus)

hannesm:
thanks @dinosaure -- I squashed some commits and force-pushed them. CI is green, merging.

hannesm:
forgot to mention here that I ran `bench/speed.exe` (which benchmarks throughput of hashes, symmetric ciphers, and fortuna speed) on nocrypto master and this PR, the results were roughly the same (minor variations)

PR #7, opened 2020-02-21
des: rename symbols to be co-linkable with nocrypto

hannesm:
so a unikernel can contain both nocrypto and mirage-crypto


hannesm:
ci is actually happy, but failed to tell github about it ;)

PR #8, opened 2020-02-21
depend on conf-pkg-config, which is executed by the cross-compilation runes

hannesm:
also mirage-xen-posix and ocaml-freestanding are depopts (and mirage-xen/mirage-solo5 are not needed at all anymore)

### DIFF
--- a/mirage-crypto.opam
+++ b/mirage-crypto.opam
@@ -13,6 +13,7 @@ build: [ ["dune" "subst"] {pinned}
          ["dune" "runtest" "-p" name "-j" jobs] {with-test} ]
 
 depends: [
+  "conf-pkg-config" {build}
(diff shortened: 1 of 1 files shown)


dinosaure:
It could be interesting to go a bit further with this PR and check with the CI installation of the package with `mirage-xen-posix` and `ocaml-freestanding`. From what I see, only freestanding is tested with FreeBSD. Can we have something like DEPOPTS with OCaml-CI? (/cc @talex5)

hannesm:
depopts with ocaml-ci is not supported as commented in https://github.com/mirage/mirage-crypto/pull/4#issuecomment-585198366

hannesm:
also, the cirrus CI (FreeBSD runners) install zarith-freestanding and thus some of that paths are covered by CI

PR #9, opened 2020-02-23
Condense the used parts of Numeric and Rng into Z_extra

hannesm:
this will require randomconv 0.1.3 to be in opam-repository (https://github.com/ocaml/opam-repository/pull/15888)


hannesm:
@dinosaure thanks for your comment, I tried to use `Z` instead of `Z_extra`, in the interface this works fine (with `include module type of Z`), in the `z.ml` file when I `include Z` and run `dune build`, I see the following:
```
File "pk/z.ml", line 3, characters 8-9:
3 | include Z
            ^
Error: Internal path Mirage_crypto_pk__Z is dangling.
       The compiled interface for module Mirage_crypto_pk__Z was not found.
```

I don't know how to solve this issue. in a OCaml toplevel, including modules of the same name works fine for me, so it looks like a `dune` issue!?

dinosaure:
@hannesm I got the same result as you and I tried to go further with `(wrapped false)`. At the end, `ocaml` can not _overload_ our `z.cm{o,x}` with the module `Z` available into `zarith.cma`. A small example can be done with:

```sh
$ echo let foo = "foo" > a.ml
$ ocamlc -c a.ml
$ ocamlc -a -o foo.cma a.cmo
$ mkdir a
$ cd a
$ echo include A > a.ml
$ echo let bar = "bar" >> a.ml
$ ocamlc -I .. -c a.ml
```

However, if we rename the second `a.ml` to `b.ml` (like `z.ml` to `z_ext.ml`), it's work:

```sh
$ mv a.ml b.ml
$ ocamlc -I .. -c b.ml
```

So again, a solution should exist but I still continue to suspect an useless complexity for nothing.

hannesm:
merging to rebase and unblock other PRs. the ocaml-ci seems to use an outdated opam-repository and will fix itself eventually. the cirrus CI only says "signaled to exit" (even after restarting the task multiple times).

PR #15, opened 2020-02-26
apply RSA-CRT hardening for PKCS1 sign operations; use eqaf for comparing cstructs in constant time

hannesm:


hannesm:
the remaining `Z.powm` occurences only use the public keys, and are thus not timing sensitive.

mirage-crypto-pk.opam
```diff
@@ -24,7 +24,7 @@ depends: [
   "ocplib-endian"
   "sexplib"
   "ppx_sexp_conv"
-  "zarith"
+  "zarith" {>= "1.4"}
```
emillon: FYI, even as of 1.9.1, `Z.powm_sec` [will fail at runtime if GMP does not support it](https://github.com/ocaml/Zarith/blob/a9a309d0596d93b6c0c902951e1cae13d661bebd/caml_z.c#L2726). You can depend on `conf-gmp-powm-sec` if want to enforce that, see https://github.com/ocaml/opam-repository/pull/7954

hannesm: thanks, done subsequently

hannesm: I was not aware of that issue and package. I was aware of the zarith code failing at runtime (and agree that failing at compile time is better) -- though if you ship a binary to some other computer, this will be dynamically linked against libgmp and fail at runtime... :/

pk/mirage_crypto_pk.mli
```diff
@@ -134,9 +134,17 @@ module Rsa : sig
...
-  (** [decrypt ~mask key ciphertext] is the decrypted [ciphertext], left-padded
-      with [0x00] up to [key] size.
+  val decrypt : ?rsa_crt_hardening:bool -> ?mask:mask -> key:priv ->
+    Cstruct.t -> Cstruct.t
+  (** [decrypt ~rsa_crt_hardening ~mask key ciphertext] is the decrypted
+      [ciphertext], left-padded with [0x00] up to [key] size.
+
+      [~rsa_crt_hardening] defaults to [false]. If [true] verifies that the
+      result is correct. This is to counter Chinese remainder theorem attacks to
+      factorize primes. If the computed signature is incorrrect, it is again
```
```suggestion
      factorize primes. If the computed signature is incorrect, it is again
```

pk/rsa.ml
```diff
@@ -77,35 +77,44 @@ and priv_bits ({ n; _ } : priv) = Numeric.Z.bits n
...
+  and m2 = Z.(powm_sec c dq q) in
   let h  = Z.(erem (q' * (m1 - m2)) p) in
-  Z.(h * q + m2)
-
-let decrypt_blinded_unsafe ?g ~key: ({ e; n; _} as key : priv) c =
+  let m  = Z.(h * q + m2) in
+  (* counter Arjen Lenstra's CRT attack by verifying the signature. Since the
+     public exponent is small, this is not very expensive. Mentioned again
+     "Factoring RSA keys with TLS Perfect Forward Secrecy" (Weimer, 2015). *)
+  if not rsa_crt_hardening || Z.(powm m e n) = c then
```
cfcs: Is there strong justification for not also just using `powm_sec` here?

hannesm: as mentioned in the initial commit, using `powm_sec` on operations where only the public parts of the asymmetric key is involved (here: `e` - the public exponent, and `n` - the public modulus) isn't really worth it, or?

hannesm:
rebased on master and force-pushed

mirage-crypto-pk.opam
cfcs: Opened https://github.com/ocaml/opam-repository/issues/15917 for discussion about this.

pk/rsa.ml
```diff
@@ -266,7 +266,7 @@ module OAEP (H : Hash.S) = struct
     let db = MGF.mask ~seed:(MGF.mask ~seed:mdb ms) mdb in
     let i  = Cs.ct_find_uint8 ~off:hlen ~f:((<>) 0x00) db |> Option.get ~def:0
     in
-    let c1 = Cs.ct_eq (sub db 0 hlen) H.(digest label)
+    let c1 = Eqaf_cstruct.equal (sub db 0 hlen) H.(digest label)
     and c2 = get_uint8 b0 0 = 0x00
     and c3 = get_uint8 db i = 0x01 in
     if c1 && c2 && c3 then Some (shift db (i + 1)) else None
```
cfcs: This style of condition checking (which is prevalent throughout `mirage-crypto`) is sometimes bad since `&&` is short-circuiting [see example](https://ocaml.godbolt.org/z/sZuy3K). Here it leads to the potential disclosure (timing-wise) of information about the outcomes of `get_uint8 b0 0 = 0x00` and `get_uint8 db i = 0x01`.

I have no idea what these values are, so I don't know if that's a problem, but it reads a bit dangerous when constant-time stuff is being used.

pk/rsa.ml
```diff
@@ -266,7 +266,7 @@ module OAEP (H : Hash.S) = struct
     let db = MGF.mask ~seed:(MGF.mask ~seed:mdb ms) mdb in
     let i  = Cs.ct_find_uint8 ~off:hlen ~f:((<>) 0x00) db |> Option.get ~def:0
```
cfcs: Is `i` as a sensitive parameter in this case?

cfcs:
Just found this, which does not sound so good: https://lists.lysator.liu.se/pipermail/nettle-bugs/2016/003104.html
> The function mpz_powm_sec() is not a drop in replacement for mpz_powm(),
although it may sound like it. If you look at the docs [1] a
requirement for mpz_powm_sec() is that the modulus must be odd.
> If the modulus is even / divisible by 2 then mpz_powm_sec() will crash
with a floating point exception.

hannesm:
I removed the usage of `powm_sec` from this PR, it looks a bit too controversial of (a) when to use (b) whether to use (failure semantics -- raises if modulus is even). Hope this PR is less controversial for @cfcs, and we can move on by merging this PR.

PR #18, opened 2020-02-26
ccm gcm adjustments

hannesm:
see individual commit messages for more details

### DIFF
--- a/src/ccm.ml
+++ b/src/ccm.ml
@@ -73,7 +73,7 @@ let gen_ctr nonce i =
   let pre, _, q = gen_ctr_prefix nonce in
   pre <+> encode_len q i
 
-let prepare_header nonce adata tlen plen =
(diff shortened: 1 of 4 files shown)


hannesm:
this changes the behaviour of GCM (now raise on empty IV), this is fine for the time being -- note that protocols on top of GCM (such as TLS - https://tools.ietf.org/html/rfc5288#section-3 where a 4 byte implicit salt (derived from KDF) and a 8 byte explicit nonce (transported in the packed) are used -- thus a bad packet (too short) will be detected before GCM.decrypt is called -- or phrased differently: any use of GCM in TLS will never provide an empty IV, but always one of 12 bytes) ensure that the IV will never be empty. there may be in the future a no-exception-leaking interface for mirage-crypto,  but that may be rather inconvenient to use.

PR #28, opened 2020-03-05
pk: make private and public key private structures

hannesm:
validate before constructing public and private keys. use result instead of exceptions in of_exp, of_prime

remove well_formed, which is now implicit.

### DIFF
--- a/pk/mirage_crypto_pk.mli
+++ b/pk/mirage_crypto_pk.mli
@@ -7,13 +7,6 @@ module Rsa : sig
 
   (** {1 Keys}
 
-      {b Warning} The behavior of functions in this module is undefined if the
-      key material is not numerically well-formed. It is the responsibility of
-      the client to ensure the trustworthiness of keys.
-
-      The two anchoring points provided are {{!generate}[generate]} and
-      {{!well_formed}[well_formed]}.
-
       Messages are checked not to exceed the key size, and this is signalled via
       the {!Insufficient_key} exception.
 
@@ -24,7 +17,7 @@ module Rsa : sig
       numerical interpretation of the (potentially padded) message is not
       smaller than the modulus. *)
 
-  type pub  = {
+  type pub = private {
     e : Z.t ; (** Public exponent *)
     n : Z.t ; (** Modulus *)
   }
@@ -32,7 +25,11 @@ module Rsa : sig
 
       {e [Sexplib] convertible}. *)
 
-  type priv = {
+  val pub : e:Z.t -> n:Z.t -> (pub, [> `Msg of string ]) result
+  (** [pub ~e ~n] validates the public key: [1 < e < n], [n > 0],
+      [is_odd n], and [numbits n >= 89] (a requirement for PKCS1 operations). *)
+
+  type priv = private {
     e  : Z.t ; (** Public exponent *)
     d  : Z.t ; (** Private exponent *)
     n  : Z.t ; (** Modulus ([p q])*)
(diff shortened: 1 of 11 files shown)


pk/mirage_crypto_pk.mli
```diff
@@ -149,16 +122,17 @@ module Rsa : sig
 
   (** {1 Key generation} *)
 
-  val generate : ?g:Mirage_crypto_rng.g -> ?e:Z.t -> bits -> priv
-  (** [generate g e bits] is a new {{!priv}private key}. The new key is
+  val generate : ?g:Mirage_crypto_rng.g -> ?e:Z.t -> bits:bits -> unit -> priv
+  (** [generate ~g ~e ~bits ()] is a new {{!priv}private key}. The new key is
       guaranteed to be {{!well_formed}well formed}.
```
```suggestion
      guaranteed to be well-formed (see {!priv} and {!pub}).
```

pk/rsa.ml
```diff
@@ -35,48 +26,147 @@ let digest_or ~hash =
...
+
+type pub = { e : Z_sexp.t ; n : Z_sexp.t } [@@deriving sexp]
+
+(* due to PKCS1 *)
+let minimum_octets = 12
+let minimum_bits = 8 * minimum_octets - 7
+
+let guard p err = if p then Ok () else Error err
+
+open Rresult.R.Infix
```
dinosaure: `open Rresult` defines infix operators. The idea is to do
```ocaml
open Rresult

(* free to use ( >>= ) and ( >>| ) and free to use [R] module. *)
let _ =
   foo x >>= bar x >>= fun () -> R.ok true
```

`Infix` module comes to mix `rresult` operator with some others like `lwt`.

pk/rsa.ml
```diff
@@ -35,48 +26,147 @@ let digest_or ~hash =
...
+  | Error (`Msg m) -> failwith "bad public key: %s" m
+
+type priv = {
+  e : Z_sexp.t ; d : Z_sexp.t ; n  : Z_sexp.t ;
+  p : Z_sexp.t ; q : Z_sexp.t ; dp : Z_sexp.t ; dq : Z_sexp.t ; q' : Z_sexp.t
+} [@@deriving sexp]
+
+let valid_prime name p =
+  guard Z.(p > zero && is_odd p && Z_extra.pseudoprime p)
+    (`Msg ("invalid prime " ^ name))
```
dinosaure: `R.msgf` is better I think.

pk/rsa.ml
```diff
@@ -35,48 +26,147 @@ let digest_or ~hash =
...
+  match priv ~e:p.e ~d:p.d ~n:p.n ~p:p.p ~q:p.q ~dp:p.dp ~dq:p.dq ~q':p.q' with
+  | Error (`Msg m) -> failwith "invalid private key %s" m
+  | Ok p -> p
+
 let priv_of_primes ~e ~p ~q =
+  valid_prime "p" p >>= fun () ->
+  valid_prime "q" q >>= fun () ->
+  guard (p <> q) (`Msg "p and q are the same prime") >>= fun () ->
   let n  = Z.(p * q)
   and d  = Z.(invert e (pred p * pred q)) in
```
cfcs: In `priv` we have
`guard Z.(one < d && d < n) (`Msg "invalid private exponent") >>= fun () ->`
Is that also guaranteed here?

pk/rsa.ml
cfcs: We're not using Lwt in this module though, so is that a problem?

dinosaure: When you `open Rresult`, you are able to use `R....`. Currently, if you want to use something like `R.msgf`, you must fully qualified `Rresult.R.msgf`. The advise comes with my other comment [to prefer `R.msgf` instead to use a concatenation](https://github.com/mirage/mirage-crypto/pull/28#discussion_r388303199).

pk/rsa.ml
hannesm: yes, indeed (from https://gmplib.org/manual/Number-Theoretic-Functions.html#Number-Theoretic-Functions):
Compute the inverse of op1 modulo op2 and put the result in rop. If the inverse exists, the return value is non-zero and rop will satisfy 0 <= rop < abs(op2) (with rop = 0 possible only when abs(op2) = 1, i.e., in the somewhat degenerate zero ring). If an inverse doesn’t exist the return value is zero and rop is undefined. The behaviour of this function is undefined when op2 is zero.

--> d will always be positive and smaller the provided modulus (here (p - 1) * (q - 1) -- which is < n).

hannesm:
I have now also adapted DSA with private structures, and validation of public and private keys. In addition, I adapted how OpenSSL blinds DSA signing, and removed a non-termination from `Dsa.sign` (now leading to an Invalid_argument - happy to hear better solutions here, see discussion in https://tools.ietf.org/html/rfc6979#section-3.4 - I believe exceptional behaviour is better than non-termination). This PR is now good to be reviewed and merged. Afterwards we can finally use `powm_sec` (which required the modulus to be odd and the exponent to be positive) without potentially triggering exceptions (happy to do this as part of this PR or in a subsequent one).

PR #30, opened 2020-03-07
mirage-crypto-pk: use powm_sec where suitable

### LINKED ISSUE #24: use powm_sec instead of powm to mitigate timing side channels
RSA optionally uses blinding for mitigation, but DH and DSA do not.

Using powm_sec should be done carefully, since it is different from powm:
- raises an exception if the exponent is negative or zero
- raises an exception if the modulus is even

there are several ways forward:
- document exceptions potentially raised
- validate input before passing it to powm_sec to avoid exceptions
- surround with try .. catch

other questions
- are blinding and powm_sec orthogonal? does it make sense to use both? (@hannesm thinks not since both attempt to mitigate the same issue)
- should `powm_sec` be used for public data (i.e. verification, c ^ e mod n), @cfcs mentions "c may be private", reference needed

hannesm:
fixes #24 //cc @cfcs @dinosaure @emillon


pk/mirage_crypto_pk.mli
```diff
@@ -1,5 +1,14 @@
 (** {1 Public-key cryptography} *)
 
+(** Public and private key types are private, the constructors validate their
+    well-formedness as much as possible, esp. so that [powm_sec] will not raise
+    an exception (exponent > 1, modulus odd). All modular exponentiations
```
```suggestion
    an exception (exponent > 1, or odd modulus). All modular exponentiations
```
clarify it's "either" here as opposed to "and" both invariants? left suggestion

pk/mirage_crypto_pk.mli
```diff
@@ -1,5 +1,14 @@
...
+    (unless otherwise noted) use the [Z.powm_sec] function, which uses a static
```
```suggestion
    (unless otherwise noted) use the {!Z.powm_sec} function, which uses a static
```

pk/mirage_crypto_pk.mli
```diff
@@ -303,7 +314,8 @@ module Dsa : sig
     (priv, [> `Msg of string ]) result
   (** [priv ~fips ~p ~q ~gg ~x ~y] constructs a private DSA key from the given
       numbers. Will result in an error if parameters are ill-formed: same as
-      {!pub}, and additionally [0 < x < q] and [y = g ^ x mod p]. *)
+      {!pub}, and additionally [0 < x < q] and [y = g ^ x mod p]. Not that no
```
```suggestion
      {!pub}, and additionally [0 < x < q] and [y = g ^ x mod p]. Note that no
```

pk/mirage_crypto_pk.mli
```diff
@@ -448,7 +461,8 @@ module Dh : sig
   (** [gen_group ~g ~bits ()] generates a random {!group} with modulus size
       [bits]. Uses a safe prime [p = 2q + 1] (with [q] prime) for the modulus
       and [2] for the generator, such that [2^q = 1 mod p].
-      Runtime is on the order of minute for 1024 bits.
+      Runtime is on the order of minute for 1024 bits. Note that no time masking
```
```suggestion
      Runtime is on the order of a minute for 1024 bits. Note that no time masking
```
Hopefully people won't be generating too many 1024-bit groups; maybe a more useful example would be for 3072-bit groups? Do we have benchmarks for those?

pk/mirage_crypto_pk.mli
cfcs: Err it is "and", sorry.

hannesm: it actually is an `or`  (the negation "not raise an exception" applies to either exponent <= 1 _or_ modulus even)

PR #32, opened 2020-03-11
Dh: revise doc/exceptions

cfcs:
Notably changes `Invalid_public_key` to `Invalid_private_key` since it was also thrown for bad secret key material when `not (valid_secret group x)`.

### DIFF
--- a/pk/mirage_crypto_pk.mli
+++ b/pk/mirage_crypto_pk.mli
@@ -413,9 +413,12 @@ module Dh : sig
 
   (** {1 Diffie-Hellman key exchange} *)
 
-  exception Invalid_public_key
-  (** Raised if the public key is degenerate. Implies either badly malfunctioning
-      DH on the other side, or an attack attempt. *)
+  exception Invalid_key
+  (** Raised if the private key material is degenerate.
+      The following invariants are checked:
+      Secret key: [1 < secret < p]
+      Public key: [1 < public < p-1] && [public <> gg]
+  *)
 
   type group = private {
     p  : Z.t ;        (** modulus *)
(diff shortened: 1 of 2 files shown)


pk/mirage_crypto_pk.mli
```diff
@@ -413,9 +413,12 @@ module Dh : sig
...
   (** {1 Diffie-Hellman key exchange} *)
 
-  exception Invalid_public_key
-  (** Raised if the public key is degenerate. Implies either badly malfunctioning
-      DH on the other side, or an attack attempt. *)
+  exception Invalid_private_key
+  (** Raised if the private key is degenerate.
+      The following invariants are checked:
+      Secret key: [1 < s < p]
+      Public key: [1 < x < p-1] && [x <> g]
```
hannesm:
```suggestion
      Public key: [1 < public < p-1] && [public <> gg]
```
the `generator` -- below in `type group` is named `gg`. I'd avoid using `x` (which is the field of the private key in `secret`)

pk/mirage_crypto_pk.mli
```diff
@@ -413,9 +413,12 @@ module Dh : sig
 
   (** {1 Diffie-Hellman key exchange} *)
 
-  exception Invalid_public_key
-  (** Raised if the public key is degenerate. Implies either badly malfunctioning
-      DH on the other side, or an attack attempt. *)
+  exception Invalid_private_key
+  (** Raised if the private key is degenerate.
+      The following invariants are checked:
+      Secret key: [1 < s < p]
```
hannesm:
```suggestion
      Secret key: [1 < secret < p]
```
`s` is a bit meaningless.

PR #33, opened 2020-03-11
remove rfc subdirectory - release artifacts are unnecessarily big (3 MB) due to these files being included

hannesm:
or is there an alternative to tell dune-release to not embed a certain directory //cc @samoht @NathanReb


avsm:
Can perhaps move these into a separate branch, as useful for reference. Merging for now.

PR #39, opened 2020-03-15
Add Windows support

avsm:
Need this to release conduit (which needs Windows support as well).  It also fixes a bug in the C stubs, and adds GH Actions CI so that we can test Windows in CI as well.

### DIFF
--- a/rng/unix/mc_getrandom_stubs.c
+++ b/rng/unix/mc_getrandom_stubs.c
@@ -42,6 +42,16 @@ void raw_getrandom (uint8_t *data, uint32_t len) {
   }
 }
 #elif (defined(__WIN32__))
+/* There is a choice between using RtlGenRandom and BCryptGenRandom
+ * here, and Microsoft does not make the choice obvious. It appears
+ * that RtlGenRandom is best used when older Windows compatibility
+ * is of concern, but requires some gymnastics around binding it
+ * with the right calling convention.
+ *
+ * Therefore (https://github.com/mirage/mirage-crypto/pull/39) we
+ * have decided to go with the more modern Windows API with bcrypt,
+ * and make Windows 10 our minimum supported version of mirage-crypto.
+ */
 #include <Windows.h>
 #include <ntstatus.h>
 #include <bcrypt.h>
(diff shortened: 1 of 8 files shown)


rng/unix/mc_getrandom_stubs.c
```diff
@@ -41,6 +41,17 @@ void raw_getrandom (uint8_t *data, uint32_t len) {
...
+#elif (defined(__WIN32__))
+#include <Windows.h>
+#include <ntstatus.h>
+#include <bcrypt.h>
+
+void raw_getrandom(uint8_t *data, uint32_t len) {
+   NTSTATUS Status;
+   Status = BCryptGenRandom(NULL, data, len, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+   if (Status != STATUS_SUCCESS)
+     uerror("BCryptRandomGen", Nothing);
```
hannesm:
```suggestion
     uerror("BCryptGenRandom", Nothing);
```

avsm:
I'm in two minds about which one to use. It's a tossup between a complex linkage (for rtlgenrandom and advapi32.dll) and unclear semantics (for BCryptGenRandom, see [here](https://stackoverflow.com/a/48876970) for the variations).

I picked the Microsoft-recommended version, but I'll give the RtlGenRandom one a shot now to see how that diff looks.  I'm only testing on Windows 10 / Server 2016.

hannesm:
@avsm thanks for your work and comments. I'm not a windows expert. I suspect if we check for a recent enough windows version, we can as well avoid using DUAL_EC_DRBG, from https://docs.microsoft.com/en-us/windows/win32/seccng/cng-algorithm-identifiers?redirectedfrom=MSDN

```
Windows 8: Beginning with Windows 8, the EC RNG algorithm supports FIPS 186-3. Keys less than or equal to 1024 bits adhere to FIPS 186-2 and keys greater than 1024 to FIPS 186-3.

Windows 10: Beginning with Windows 10, the dual elliptic curve random number generator algorithm has been removed. Existing uses of this algorithm will continue to work; however, the random number generator is based on the AES counter mode specified in the NIST SP 800-90 standard. New code should use BCRYPT_RNG_ALGORITHM, and it is recommended that existing code be changed to use BCRYPT_RNG_ALGORITHM. 
```

-> I'd be happy with a code comment (why this is fine) and a test (if achievable) that at least windows 10 is used

avsm:
Decision with @hannesm: we will use bcrypt as its more modern, and Windows 10 is our base of support anyway. I'll push a README and code comment note to this effect.

avsm:
This should be ready for merge now. I rearranged the doc slightly in 2260884 @hannesm, if you want to glance over that.

hannesm:
ok, that looks good. could you add a `dune-configurator {build}` dependency to `mirage-crypto-rng.opam`, please?

avsm:
Good catch! Done and will merge when CI passes

PR #41, opened 2020-03-18
Provide a new implementatio of `ct_find_uint8`

dinosaure:
This is a draft about a constant-time implementation of `ct_find_uint8` according @cfcs's proposal.

### DIFF
--- a/pk/rsa.ml
+++ b/pk/rsa.ml
@@ -235,7 +235,9 @@ module PKCS1 = struct
 
   let unpad ~mark ~is_pad cs =
     let f = not &. is_pad in
-    let i = Cs.ct_find_uint8 ~off:2 ~f cs |> Option.get ~def:2 in
+    let i =
+      let v = Cs.ct_find_uint8 ~off:2 ~f cs in
+      if v = (-1) then 2 else v in
     let c1 = get_uint8 cs 0 = 0x00
     and c2 = get_uint8 cs 1 = mark
     and c3 = get_uint8 cs i = 0x00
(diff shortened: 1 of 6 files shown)


pk/rsa.ml
```diff
@@ -338,8 +340,9 @@ module OAEP (H : Hash.S) = struct
   let eme_oaep_decode ?(label = Cstruct.empty) msg =
     let (b0, ms, mdb) = Cs.split3 msg 1 hlen in
     let db = MGF.mask ~seed:(MGF.mask ~seed:mdb ms) mdb in
-    let i  = Cs.ct_find_uint8 ~off:hlen ~f:((<>) 0x00) db |> Option.get ~def:0
-    in
+    let i  =
+      let v = Cs.ct_find_uint8 ~off:hlen ~f:((<>) 0x00) db in
+      if v = (-1) then 0 else v in
```
cfcs: `Eqaf.select` here

cfcs:
@dinosaure isn't this a utility function that we should provide in `eqaf` ? It seems useful to other libraries too, and I think it would be better to maintain the constant-time things that are not crypto-related in `eqaf` instead of `mirage-crypto`

pk/rsa.ml
dinosaure: How you want to use `select_int` here? I mean the condition here is `(-1)` and `choose_b` according to the behavior of `select_int` must be `0` or something else.

dinosaure:
> @dinosaure isn't this a utility function that we should provide in eqaf ? It seems useful to other libraries too, and I think it would be better to maintain the constant-time things that are not crypto-related in eqaf instead of mirage-crypto

I agree with you but I did a release of `eqaf` few days ago. May be you can have an iterative plan where `ct_find_uint8` still is into `mirage-crypto` for the next release of it and prepare `eqaf` in consequence?

hannesm:
I'd prefer that function to be in eqaf, tbh. once we have it in there (and released), we can use it in mirage-crypto. thanks.

dinosaure:
> I'd prefer that function to be in eqaf, tbh. once we have it in there (and released), we can use it in mirage-crypto. thanks.

Ok,, I will post-pone that next week :+1:.

PR #44, opened 2020-03-25
Update test.yml

imbsky:
Specifying action version by a branch name instead of a git tag is dangerous because it can break a build at an unintended time. Ideally, it should be specified by a commit hash, but let's use the common method (git tag) for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,28 +1,39 @@
 name: Crypto
+
 on: [push, pull_request]
+
 jobs:
   tests:
     name: Tests
-    runs-on: ${{ matrix.operating-system }}
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


hannesm:
would you mind to elaborate why using "@v2.0.0"  and "@v1.0" is a good idea? will this break at a point where mirage-crypto `v2.0.0` gets tagged? (sorry, I just don't know much about these actions, but using `master` as the de-facto main branch name looks fine with me)

avsm:
@hannesm this isn't to do with the version of the code being checked out; it's the version of the GitHub Action that is being invoked. The PR locks down the versions of those actions to their respective git tags, so commits upstream to the Actions won't change our CI.  We need to manually bump this whenever we want to take advantage of new features in the relevant Actions, but that's fine.

Issue #46, opened 2020-04-06
Use dune's variants and trick on the META file


dinosaure:
This issue wants to solve an initial and systematic issue found into `checkseum`/`digestif` and try to prepare an easily move to `mirage` with `dune`. To contextualize a bit the goal of this issue, you probably should look into mirage/mirage#1024.

Come back to the _dunification_ of MirageOS, the main and fully described difference of the compilation of an _unikernel_ with `ocamlbuild` and with `dune` is the option: `-output-complete-obj`. The main difference is about static C libraries. Let back a bit about the diff.

#### Link an _unikernel_ with `ocamlbuild` + `ocamlfind`

According the current status of MirageOS, a library with C stubs such as `checkseum` must fill a `META` file with some _metadata_ to let the `mirage` tool to choose the right static C library according the target. `mirage` asks with `ocamlfind` these _metadata_ and put at the link-time with `-output-obj` static libraries collected.

The compiler, at the link time and **with** `-output-obj` does not care about static libraries described into `*.cmxa`. This is why the `mirage` tool/`ocamlbuild`/`ocaml` **can** chooses the right static C libraries such as `-lgmp-freestanding` (for Solo5) according the target (even if `zarith.cmxa` describes `-lgmp`).

#### Link an _unikernel_ with `dune`

However, `dune` strictly uses a new option of the compiler, `-output-complete-obj` and it seems that `-output-obj` will be deprecated. This new option change the behavior of the compiler where it wants to link the final `main.o` (the _unikernel_) with libraries described into `*.cmxa` used to produce the _unikernel_.

A link with `zarith.cmxa` with `-output-complete-obj` will try (at the compiler layer) to link with `libgmp.a` in any cases.

#### For a library such as `checkseum` (and `mirage-crypto`)

Of course, for any libraries which wants to provide C stubs, a compilation of the _unikernel_ with `-output-complete-obj` constraints us to:
1) still continue to provide a `META` file with such _metadata_ to be compatible with MirageOS 3.*
2) provides well-formed `*.cmxa` with right C libraries such as:
    + `checkseum_freestanding.cmxa` should link with `-lcheckseum_freestanding`
    + `checkseum_xen.cmxa` should link with `-lcheckseum_xen`

The second point is a requirement of the compiler due to the use of `-output-complete-obj`. The first point is the well-know requirement for MirageOS 3.*.

[...]

PR #47, opened 2020-04-06
Fix C flags to allow cross-compilation

samoht:
/cc @dinosaure and @avsm (who helped me debugging the issue)

### DIFF
--- a/freestanding/dune
+++ b/freestanding/dune
@@ -5,7 +5,7 @@
  (public_name mirage-crypto.freestanding)
  (optional)
  (libraries ocaml-freestanding)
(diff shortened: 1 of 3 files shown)


dinosaure:
Currenlty with `dune.2.4.0`, `:standard` puts `-O2 -fno-strict-aliasing -fwrapv -fPIC`, you can check it with:
```sh
$ dune printenv
```

hannesm:
thanks, as noted in #46 please let us wait until I conducted some experiments before merging. thanks.

hannesm:
should the same `(:standard)` be added in the `freestanding` and `xen` subdirectories? would be good to keep them in sync.

hannesm:
thanks

PR #48, opened 2020-04-07
Variants

dinosaure:
A proposal about #46

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,17 +1,15 @@
 (library
   (name mirage_crypto)
   (public_name mirage-crypto)
-  (libraries cstruct ocplib-endian)
-  (private_modules ccm cipher_block cipher_stream hash native uncommon)
-  (c_names misc
-           md5 sha1 sha256 sha512 hash_stubs
-           aes_generic aes_aesni ghash_generic ghash_pclmul
-           des_generic
-           entropy_cpu_stubs)
-  (c_flags (:standard) (:include cflags.sexp)))
+  (modules mirage_crypto ccm cipher_block cipher_stream hash uncommon)
+  (libraries mirage_crypto_native cstruct ocplib-endian)
+  (private_modules ccm cipher_block cipher_stream hash uncommon))
+
+(library
+  (name mirage_crypto_native)
+  (public_name mirage-crypto.native)
+  (virtual_modules native)
+  (modules native)
+  (default_implementation mirage-crypto.native.unix))
 
 (include_subdirs unqualified)
-
-(rule
-  (deps (env_var MIRAGE_CRYPTO_ACCELERATE))
-  (action (with-stdout-to cflags.sexp (run ../config/cfg.exe))))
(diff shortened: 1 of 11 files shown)


src/dune
```diff
@@ -1,17 +1,15 @@
...
   (public_name mirage-crypto)
-  (libraries cstruct ocplib-endian)
-  (private_modules ccm cipher_block cipher_stream hash native uncommon)
-  (c_names misc
-           md5 sha1 sha256 sha512 hash_stubs
-           aes_generic aes_aesni ghash_generic ghash_pclmul
-           des_generic
-           entropy_cpu_stubs)
-  (c_flags (:standard) (:include cflags.sexp)))
+  (modules mirage_crypto ccm cipher_block cipher_stream hash uncommon)
```
dinosaure, line 4: Currently, if I don't explicitly __all__ modules into this field (even if some appear into `private_modules`), I got a compilation error. Not sure about implication of it.

hannesm:
I don't quite understand why a `native.mli` is needed now. before this PR, the cross-compilation created an (empty!) `mirage_crypto_freestanding.ml` (and compilation resulted in cma/cmi/cmt/cmx/cmxa/cmxs thereof). can't we similarly use an empty mli, generated on-the-fly by dune somehow, and declare this as variant (implementation only matters in terms of C code anyways)?

dinosaure:
> I don't quite understand why a native.mli is needed now.

This is needed for variants/virtual library where a common (needed) interfaces can hide several implementations (unix, xen and freestanding). So, `native.mli` is required to be used by others modules (such as `uncommon.ml` or `hash.ml`) even if the implementation of it does not exist.

> the cross-compilation created an (empty!) mirage_crypto_freestanding.ml (and compilation resulted in cma/cmi/cmt/cmx/cmxa/cmxs thereof)

`mirage_crypto_freestanding` was empty because we just wanted to produce a `libmirage_crypto_freestanding_stubs.a`. Note that we don't want to link or use it but just generate the needed artifact for a Solo5 _unikernel_.

Then, the current `mirage` tool will take care about the link and will use `libmirage_crypto_freestanding.a` or `libmirage_crypto_xen.a` according the target - again, without an introspection of needed static C library notified into `*.cmxa` (even if they are empty). The current status of `mirage-crypto` generates a `mirage_crypto.cmxa` with `-lmirage_crypto_stubs` which corresponds to a _standalone_ compilation (incompatible with Xen and Solo5) - again, this wrong behavior appears only with `-output-complete-obj`.

The idea is to use a sub-library `mirage_crypto_native` which does not refer to any implementations (unix, xen or freestanding) and let `dune` (with _variants_) to choose one of them at the end.

> use an empty mli, generated on-the-fly by dune somehow, and declare this as variant (implementation only matters in terms of C code anyways)?

The problem is that `native` does not come with any implementations (`mirage_crypto_native` does have any `*.cmx`). So, we need something (a `*.mli`) to say to others that this module _should_ provides these functions. By this way, they are available but does not exists yet - you need a link to `mirage-crypto.native.{unix,freestanding,xen}`.

hannesm:
should this be part of the next release (sorry, I lost track how mirage4 plans to handle C resources). if it is the case, would you mind to rebase this on top of master and look into the CI failures (the windows runner has some issues).

hannesm:
dune removed their variants, the MirageOS 4 story is laid out slightly different now. closing this PR.

PR #49, opened 2020-04-13
Uncommon: removed Cs.create (use Cstruct.create and Cstruct.memset instead)

hannesm:
Uncommon: Cs.clone removed unused ?off parameter

### DIFF
--- a/rng/hmac_drbg.ml
+++ b/rng/hmac_drbg.ml
@@ -12,8 +12,11 @@ module Make (H : Mirage_crypto.Hash.S) = struct
 
   let (bx00, bx01) = Cs.(b 0x00, b 0x01)
 
-  let k0 = Cs.create ~init:0x00 H.digest_size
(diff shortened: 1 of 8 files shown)


rng/hmac_drbg.ml
```diff
@@ -12,8 +12,11 @@ module Make (H : Mirage_crypto.Hash.S) = struct
 
   let (bx00, bx01) = Cs.(b 0x00, b 0x01)
 
-  let k0 = Cs.create ~init:0x00 H.digest_size
-  and v0 = Cs.create ~init:0x01 H.digest_size
+  let k0 = Cstruct.create H.digest_size
```
dinosaure, line 15: Hmmhmm, it's not clear that `Cstruct.create` do properly a `memset`. It should be better to explicit this behavior with a comment I think.

hannesm: done

PR #51, opened 2020-04-16
mirage-crypto: remove ocplib-endian dependency, use Bytes.set_int64 in Counters (also remove Uncommon.Option in favour of stdlib)

hannesm:
to shrink the dependency cone

### DIFF
--- a/mirage-crypto-entropy.opam
+++ b/mirage-crypto-entropy.opam
@@ -14,7 +14,7 @@ build: [
 ]
 depends: [
   "dune" {>= "1.7.0"}
-  "ocaml" {>= "4.07.0"}
+  "ocaml" {>= "4.08.0"}
   "cstruct" {>= "4.0.0"}
   "mirage-runtime" {>= "3.7.0"}
   "lwt" {>= "4.0.0"}
--- a/mirage-crypto-pk.opam
+++ b/mirage-crypto-pk.opam
@@ -14,7 +14,7 @@ build: [ ["dune" "subst"] {pinned}
 
 depends: [
   "conf-gmp-powm-sec" {build}
-  "ocaml" {>= "4.07.0"}
+  "ocaml" {>= "4.08.0"}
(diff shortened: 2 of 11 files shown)


dinosaure:
Can we use `caml_bytes_set64` instead to have the bound checking?

hannesm:
I'm even tempted to just bump the minimal OCaml version to 4.08 and use `Bytes.set_int64_be` directly. I only hesitate doing so because somehow our mirage-xen backend only supports 4.07 and 4.08 AFAICT (and my recent cleanups https://github.com/mirage/ocaml-freestanding/pull/65 (to fix https://github.com/mirage/ocaml-freestanding/issues/63 / https://github.com/mirage/ocaml-freestanding/issues/70) highlight that copy + paste cases in mirage-xen-ocaml is not the way to move forward (since mirage-xen-ocaml is pre-autoconf, and dangerously copies around header files which changed since then).

OTOH, the intersection of 4.08 and 4.07, 4.08 is non-empty. I guess we really badly need the mirage-xen overhaul planned since years...

hannesm:
@dinosaure interesting. using the bounds-checked `set` here has some performance penalty (in my unscientific tests it is <1%), and the `set` operation is always safe (there's a pre-allocated `let _tmp = Bytes.create 32^W16`, the offsets which are written are statically known (either 0 or 8)). but given that `set64u` is not exposed by the stdlib, I feel it's not worth to complicate the code and/or build system.

hannesm:
turns out, with 4.08 we can remove the Uncommon.Option and use Stdlib.Option instead \o/

hannesm:
HEADSUP this raises the lower bound to 4.08. has anyone @mirage/core strong opinions why to keep 4.07 compatibility? merge plan is early next week (April 20th).

avsm:
4.08 and higher is fine.

hannesm:
thanks, merging. I like the smell of removing code :)

Issue #23, opened 2020-02-28
Uncommon.Cs.ct_find_uint8 is not constant-time


cfcs:
The code currently looks like this (I've inserted comments):
```ocaml
  let ct_find_uint8
    ?(off=0)  (* branch on value of [off: int option] *)
    ~f cs =
    let rec go acc i = function
      | 0 -> acc
      | n ->
          let acc = match (acc, f (get_uint8 cs i))
           with 
(* [match] introduces a branch; since we branch on a sensitive value this
means we leak timing of whether or not [f] is true.
Example: If we always take one branch, the branch predictor will be
  hot for our code.
If we find something in the middle, we disrupt that.
Depending on at what offset we find a match, we may or may not prime
 the branch predictor again. That's worse than just leaking the predicate,
since it can leak the index of the the matching predicate.

             *)
            | (None, true) ->

              Some i (* allocating a new Some box is observable,
                     so we leak the information that predicate [f] is true *)
            | _            -> acc in
          go acc (succ i) (pred n) in
    go None off
    (len cs - off) (* if [off] is > len cs the loop never terminates *)
```

@dinosaure Should we write a constant-time version of this for `eqaf`? (e.g. take non-optional offset, always update `acc` using constant-time `select` (maybe we should also expose our constant-time `select` code)?

EDIT(s): Formatting + elaboration

dinosaure:
`eqaf` was just released into ocaml/opam-repository#15982 with `select_int`, `one_if_not_zero` and `zero_if_not_zero` 🎉

hannesm:
@dinosaure cool, once that is in opam-repository, feel free to PR here changes to use it accordingly.

hannesm:
this has been addressed by #52, thanks.

PR #52, opened 2020-04-17
Replace ct_find_uint8 by Eqaf_cstruct.find_uint8

dinosaure:
A fix of #23

### DIFF
--- a/mirage-crypto-pk.opam
+++ b/mirage-crypto-pk.opam
@@ -24,7 +24,7 @@ depends: [
   "sexplib"
   "ppx_sexp_conv"
   "zarith" {>= "1.4"}
-  "eqaf" {>= "0.5"}
+  "eqaf" {>= "0.7"}
   "rresult" {>= "0.6.0"}
   ("mirage-no-xen" | "zarith-xen")
   ("mirage-no-solo5" | "zarith-freestanding")
--- a/pk/rsa.ml
+++ b/pk/rsa.ml
@@ -7,6 +7,11 @@ open Common
 let two = Z.(~$2)
 and three = Z.(~$3)
 
+(* A constant-time [find_uint8] with a default value. *)
+let ct_find_uint8 ~default ?off ~f cs =
+  let res = Eqaf_cstruct.find_uint8 ?off ~f cs in
+  Eqaf.select_int (res + 1) default res
+
 let (&.) f g = fun h -> f (g h)
 
 module Hash = Mirage_crypto.Hash
@@ -235,7 +240,7 @@ module PKCS1 = struct
 
   let unpad ~mark ~is_pad cs =
     let f = not &. is_pad in
-    let i = Cs.ct_find_uint8 ~off:2 ~f cs |> Option.get ~def:2 in
+    let i = ct_find_uint8 ~default:2 ~off:2 ~f cs in
     let c1 = get_uint8 cs 0 = 0x00
     and c2 = get_uint8 cs 1 = mark
     and c3 = get_uint8 cs i = 0x00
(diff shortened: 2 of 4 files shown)


hannesm:
this is fine, though:
- CI fails (would you mind to look into this and fix it?)
- can we remove the ct_find funcion from Uncommon, and have it in pk/common.ml or pk/rsa.ml (the only use-side)? <- I'm not keen to introduce dependencies in the main library

hannesm:
ah, the test failure (from https://pipelines.actions.githubusercontent.com/qoofUrGaI3vjbXV0l1nsaHtqOWUOy1R9qZ8Weal5PXKQiJlyok/_apis/pipelines/1/runs/28/signedlogcontent/11?urlExpires=2020-04-17T11%3A04%3A42.5829588Z&urlSigningMethod=HMACV1&urlSignature=yynvhHhYHHE%2Bp0YvoUktZOIeTSsKr2EZL95Ynsr2r%2B8%3D):
```
2020-04-17T10:40:09.1333537Z The following dependencies couldn't be met:
2020-04-17T10:40:09.1334251Z   - mirage-crypto -> eqaf >= 0.7
2020-04-17T10:40:09.1334503Z       no matching version
```

so, this will eventually solve itself (once the windows runner has a fresher opam repository!?)

dinosaure:
> can we remove the ct_find funcion from Uncommon, and have it in pk/common.ml or pk/rsa.ml (the only use-side)? <- I'm not keen to introduce dependencies in the main library

It's done :+1:.

hannesm:
thanks again, sorry for being picky, but I'd appreciate a function definition (a la `let ct_find_uint8 ..`) and avoid repeating the two lines multiple times :)

hannesm:
hmm, or not - I'm not sure anymore. why should `select_int` be used instead of `Option.get`?

dinosaure:
> hmm, or not - I'm not sure anymore. why should select_int be used instead of Option.get?

`select_int` is checked as a `constant-time` way to simulate a branch (like `if a then b else c`). I use it to replace `Option.get` which needs an allocation (and may be delay the computation). Finally, we an _assert_ that:

```ocaml
let res = Eqaf.find_uint8 ... in
let v = Eqaf.select_int (res - 1) default res in
```

Is _constant-time_ where:

```
let res = Eqaf.find_uint8 ... in
let res = if res = (-1) then None else Some res in
let v = Option.get ~def res in
```

is not. Instead, we can implement a `ct_find_uint8 : default:int -> ?off:int -> f:(int -> bool) -> Cstruct.t -> int` which uses `select_int` - then, it will be a one-liner where `find_uint8` is used.

hannesm:
@dinosaure thanks for your answer. yes, I'd appreciate a `find_uint8` that takes a `default:` argument instead of returning `-1` (which is slightly surprising in a typed OCaml world). maybe worth to do and cut a 0.7.1 or 0.8.0 release of eqaf?

hannesm:
oh, another way to move here is: develop such a `find_uint8 ~default:` locally in `rsa` and use this at the three locations. :)

dinosaure:
> returning -1 (which is slightly surprising in a typed OCaml world)

Yes, it's mostly due to the limitation to avoid any call to the GC in your `eqaf`'s codebase.

> maybe worth to do and cut a 0.7.1 or 0.8.0 release of eqaf?

Sure, may be today, or this week :+1:.

dinosaure:
> oh, another way to move here is: develop such a find_uint8 ~default: locally in rsa and use this at the three locations. :)

Done in this way :+1: into my last commit. Should I rebase?

hannesm:
thanks, merged. the CI failures (on windows) are lacking the eqaf 0.7 release in [opam-repository-mingw](https://github.com/fdopen/opam-repository-mingw/) (that will eventually appear at some point) ;the OCaml-CI has an update cycle of "around a week" (there's discussion on https://github.com/ocurrent/ocaml-ci/issues/26 about ways out -- no, embedding submodules into this repository won't happen, sorry).

since we use a variety of CI systems, we still have some that test this PR successfully :)

PR #54, opened 2020-04-28
no bitfn anymore

hannesm:
this removes code \o/

more specifically, it copes with (a) ocaml-freestanding getting an `endian.h` (https://github.com/mirage/ocaml-freestanding/pull/73) and (b) solo5 not copying endian.h around on BSD systems (https://github.com/Solo5/solo5/pull/453)


hannesm:
what is not clear to me whether this breaks e.g. windows (or any other target we're not explicitly testing in CI atm, but which is important to keep working)

hannesm:
CI fails on alpine - I thought they have musl and their `endian.h`, which defines these bindings (htole64 etc.) -- anyone has a clue what is wrong?

PR #55, opened 2020-04-30
travis: test on xenial and bionic, test on arm64

hannesm:
the macos runner has been removed since GitHub actions are executed on that platform


hannesm:
merging: this is just a CI change, and it works and helps to evaluate future PRs.

PR #56, opened 2020-04-30
Bitfn cleanups

hannesm:
this is an attempt to tidy up "bitfn.h", see individual commits for changes.


hannesm:
turns out no actual need for `<endian.h>`, we can use the compiler defined `__BYTE_ORDER__` (and `__builtin_bswapYY`. anything speaks against that?

src/native/bitfn.h
```diff
@@ -26,160 +26,60 @@
...
-}
+#define bitfn_swap32(x) __builtin_bswap32(x)
+#define bitfn_swap64(x) __builtin_bswap64(x)
 
-static inline uint64_t bitfn_swap64(uint64_t a)
-{
-	__asm__ ("bswap %0" : "=r" (a) : "0" (a));
-	return a;
-}
+#if defined(_MSC_VER)
```
avsm, line 32: I think we're not actually testing this in CI at the moment, as we're using Cygwin on Windows. I'll do a manual test later, but not a blocker for merge I think.

hannesm: I indeed did not test this, but took it from https://gist.github.com/jtbr/7a43e6281e6cca353b33ee501421860c#file-endianness-h-L67-L70

I am also not sure whether the windows native C compiler defines `__BYTE_ORDER__` and `__builtin_swapYY`.

It would be appreciated if you test whether the code works there (and compare whether the old version worked there). If we can get native windows into CI, this would as well be appreciated.

PR #58, opened 2020-05-03
bitfn: remove MSC support (untested, likely incomplete, not in CI)

hannesm:
/cc @avsm IMHO better to remove the code for now, and once we have CI and the need for a native windows port, we can work on that.


avsm:
Lgtm

PR #59, opened 2020-05-03
in development mode, add -Werror to CFLAGS

hannesm:
thanks to @samoht for the hint (I tested this locally, and a `dune build --verbose` showed me `-Werror`, a `dune build --profile=release --verbose` not -- I'm not sure whether it is correct, though)


hannesm:
I tested this change in CI (by adding a commit which leads to a C warning). The CI systems (ocaml-ci, cirrus, GitHub actions) failed (due to -Werror); while travis succeeded. This was at first surprising for me, but after investigating how the CI is setup it turns out, travis uses `opam pin` and thus is in "release mode" - while the other systems call "dune build" directly.

PR #63, opened 2020-05-12
RSA: accept keys where d is not valid (but dp and dq are fine): they occur in the wild

hannesm:


pk/rsa.ml
```diff
@@ -91,8 +91,15 @@ let priv ~e ~d ~n ~p ~q ~dp ~dq ~q' =
...
-    (`Msg "d <> e ^ -1 mod (p - 1) * (q - 1)") >>= fun () ->
+  guard Z.(one = dp * e mod (pred p)) (`Msg "1 <> dp * e mod (p - 1)") >>= fun () ->
+  guard Z.(one = dq * e mod (pred q)) (`Msg "1 <> dq * e mod (q - 1)") >>= fun () ->
+  let d =
+    let d' = Z.(invert e (pred p * pred q)) in
+    if not (d = d') then
+      Log.warn (fun m -> m "invalid d %a in private key, adjusting to %a"
+                   Z.pp_print d Z.pp_print d');
+    d'
+  in
```
hannesm, line 102: the reason to adjust the value `d` here is because we actually use `d` in decrypt if crt_check is enabled and fails (as a fallback) -- constructing private keys where `d` is wrong feels like a bad choice (may lead to issues further down the line). (input welcome @cfcs @avsm @dinosaure @emillon)

dinosaure:
I'm not an expert of crypto things but should we return an error where `d <> d'`?

hannesm:
@dinosaure that is the current behaviour, see #62 for details why this is problematic -- `dp` and `dq` (derived from `d`) are correct and actually used, `d` is incorrect but not used.

pk/rsa.ml
emillon: I agree that building an inconsistent key is a problem. But in that case, I don't think we have enough information in mirage-crypto to determine the correct solution. Maybe p, q and the CRT params are "wrong" but d is right? I think it's better to be safe and return an error.

hannesm: Thanks for your remark. This PR stem from the report in #62, and my point of view is:
- we take (and verify) p (being probabilistic prime), q (probabilistic prime), n (= p * q), e (coprime to p - 1 & q - 1)
- about d, dp, dq: we used to ensure d = inverse e ((p - 1) (q - 1)); dp = d mod p; dq = d mod q

now we relax to dp * e mod (p - 1) = 1 && dq * e mod (q - 1) = 1 for pragmatic reasons: all RSA implementation I could find use dp and dq for decryption / computing the signature -- d is mostly irrelevant (apart from our "CRT hardening" branch where if `m ^ e mod n <> c`, we compute `m = c ^ d mod n`).

we could as well attempt to recover the right set of d, dp, dq:
- if d <> inverse e ((p - 1) (q - 1)), and dp * e = 1 mod (p - 1) and dq * e = 1 mod (q - 1), use dp, dq, re-compute d
- if d = inverse e ((p - 1) (q - 1)), but dp * e <> 1 mod (p - 1) and/or dq * e <> 1 mod (q - 1), use d and recompute dp & dq

we could as well remove that fallback computation from `decrypt_unsafe` (and fail in decrypt if `m ^ e mod n <> c`) - which would lead to either using a result type, or raising an exception in such a case -- if doing so, I would not see a reason to check d at all (since we would never use that parameter).

TL;DR: such private keys are generated out there, and other tools are coping with them, what is the reason for us to reject them? OTOH, we could adapt ocaml-x509 to fall back to `Rsa.priv_of_primes` and only use p, q, e from the private key in the PEM.

hannesm: @emilon I see your point, and by now think the "fix" should be in X509, to avoid questionable changes to the non-broken mirage-crypto (which already provides functionality for recovering from "partial" private keys).

emillon: Yes this seems to be the right fix to me!

hannesm:
superseeded by https://github.com/mirleft/ocaml-x509/pull/135

PR #69, opened 2020-06-17
Mirage_crypto_rng_lwt.initialize now returns unit instead of unit Lwt.t

hannesm:
Calling Lwt.async and Lwt_unix.Enter_iter_hooks.add_first does not need to be
inside of the Lwt_main.run. This allows library code to execute
Mirage_crypto_rng_lwt.initialize () at top level.

Initialization is only done once (the `running` (bool ref) is used to figure out
whether initialize was already called before).

### DIFF
--- a/tests/test_entropy.ml
+++ b/tests/test_entropy.ml
@@ -12,9 +12,10 @@ module Printing_rng = struct
   let reseed ~g:_ data =
     Format.printf "reseeding: %a@.%!" Cstruct.hexdump_pp data
 
-  let accumulate ~g:_ ~source =
+  let accumulate ~g:_ source =
     let print data =
-      Format.printf "accumulate: (src:%d) %a@.%!" source Cstruct.hexdump_pp data
+      Format.printf "accumulate: (src: %a) %a@.%!"
+        Mirage_crypto_rng.Entropy.pp_source source Cstruct.hexdump_pp data
     in
     `Acc print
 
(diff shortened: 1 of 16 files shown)


hannesm:
as far as I'm concerned this should satisfy the conduit-lwt-unix use case (//cc @avsm), and also splits the mirage-crypto-rng-mirage package out (//cc @samoht).

tests/test_entropy.ml
```diff
@@ -12,9 +12,9 @@ module Printing_rng = struct
   let reseed ~g:_ data =
     Format.printf "reseeding: %a@.%!" Cstruct.hexdump_pp data
 
-  let accumulate ~g:_ ~source =
+  let accumulate ~g:_ source =
     let print data =
-      Format.printf "accumulate: (src:%d) %a@.%!" source Cstruct.hexdump_pp data
+      Format.printf "accumulate: (src:%d) %a@.%!" (Obj.magic source) Cstruct.hexdump_pp data
```
avsm, line 17: even in a test case, the Obj makes me sweat a little :-) How about just defining a `source_pp` function to pretty print it?  There is already a `pp_sources` above.

hannesm: thanks for your remark, I reverted some of these changes, and now there's an Obj-free implementation. a source is now (internally) a pair of int and string -- makes printing and extraction much easier :)

I force-pushed to this PR

hannesm:
Failing checks:
- github action since it fails to download some tarballs (exit code 40), other runners succeed (and I can't find a restart button)
- cirrus due to new package mirage-crypto-rng-mirage which opam-version-numbering magic gets in the way (`opam install --deps-only .` is used, but that package is not in opam-repository -> version is considered `dev`, and it has a dependency on `mirage-crypto-rng {= version}`) --> I'm too lazy to fix this (works fine locally on a FreeBSD 12)

PR #70, opened 2020-06-19
mirage-crypto-rng-mirage: require mirage-runtime 3.8.0 to avoid broken installations, mirage-crypto-rng: conflict with mirage-runtime <3.8.0

hannesm:
what happened:
- mirage outputs a dependency (when using the random device) on mirage-crypto-rng ~min:"0.7".
- it uses Mirage_crypto_rng_mirage.initialize () (from the mirage-crypto-rng.mirage sublibrary)

now, that sublibrary is gone in 0.8, and part of the separate mirage-crypto-rng-mirage package. since existing mirage tools do not emit an upper bound for the dependency, we need to ensure that no 3.7 unikernel will use mirage-crypto-rng.0.8 (otherwise a link failure will occur that mirage-crypto-rng.mirage does not exist).

### DIFF
--- a/mirage-crypto-rng.opam
+++ b/mirage-crypto-rng.opam
@@ -26,6 +26,7 @@ depends: [
   "mtime"
   "lwt" {>= "4.0.0"}
 ]
+conflicts: [ "mirage-runtime" {< "3.8.0"} ]
 description: """
 Mirage-crypto-rng provides a random number generator interface, and
 implementations: Fortuna, HMAC-DRBG, getrandom/getentropy based (in the unix
--- a/mirage-crypto-rng-mirage.opam
+++ b/mirage-crypto-rng-mirage.opam
@@ -18,7 +18,7 @@ depends: [
   "mirage-crypto-rng" {=version}
   "duration"
   "cstruct" {>= "4.0.0"}
-  "mirage-runtime" {>= "3.7.0"}
+  "mirage-runtime" {>= "3.8.0"}
   "mirage-time" {>= "2.0.0"}
   "mirage-clock" {>= "3.0.0"}
   "mirage-unix" {with-test & >= "3.0.0"}


mirage-crypto-rng.opam
```diff
@@ -26,6 +26,7 @@ depends: [
   "mtime"
   "lwt" {>= "4.0.0"}
 ]
+conflicts: [ "mirage-runtime" {< "3.8.0"} ]
```
samoht, line 29: do you really need this conflict? e.g. why isn't the other diff line not enough?

hannesm: mirage 3.7.x emits a dependency onto (a) mirage-runtime `>= 3.7 & < 3.8` (b) (if random is used) mirage-crypto-rng `>= 0.7`.

Now, in mirage 3.8 this got fixed to mirage-crypto-rng-mirage `>= 0.8 & < 0.9`. but I expect there are users with mirage 3.7 out there, and their opam will happily install mirage-crypto-rng 0.8 otherwise (and then they'll get a build-failure since that one doesn't provide a `.mirage` sublibrary / Mirage_crypto_rng_mirage module).

The issue was that I forgot to put an upper bound in the mirage tool for the mirage-crypto-rng dependency; and then someone kindly asked to move the mirage specific stuff into a separate opam package.

Other options I could think of:
- have the conflict in mirage-runtime 3.7.x packages (onto mirage-crypto >= 0.8)
- provide a backwards-compatible "mirage-crypto-rng.mirage" ocamlfind library which says "use mirage-crypto-rng-mirage instead" (not sure whether this would help -- the user being confused since the dependency is added by the mirage tool)
- your option here

the good news is that once we move forward (i.e. mirage 4) we can just clean up this conflict and no further support the 3.x series :)

the actual solution is to not embed version and package information in the mirage tool, but instead in e.g. opam metadata (see https://github.com/roburio/mirage-noconfig/issues/9#issuecomment-549145519 for some ideas, I'm fine to use some other place for this metadata, but imho it should be in the package information rather than in the mirage tool).

hannesm:
merging, CI failure is windows opam-repository lagging a bit behind

PR #72, opened 2020-06-26
Add Chacha20 cipher, and poly1305 mac

hannesm:
Differences include:
 - OCaml API (no separate expand and encrypt/decrypt),
 - support for IETF mode (nonce 12 bytes, counter 32bit)
 - reduced allocation overhead


hannesm:
A second commit adds the poly1305 mac. Another extension (AEAD construction of poy1305/chacha20) is on its way. I'm curious about feedback in respect to the 1305 interface.

hannesm:
that's it, again ensured manually that 036c143 compiles and works on a 32bit switch.

PR #73, opened 2020-06-26
Unify AEAD constructions

hannesm:


hannesm:
this is rebased, and includes a module type AEAD now as well (which CCM, GCM, and Chacha20_poly1305 adhere to). comments welcome.

hannesm:
questions about this approach:
- it would be nice to include the `of_secret : Cstruct.t -> key` into the module type, but then CCM's `of_secret` also takes a `maclen:int`, should this be revised (if so, any hints how?)
- the `of_secret` can raise if the maclen / key is not of appropriate size, this is fine I guess
- furthermore the decrypt and encrypt operations may raise if the provided `nonce` is not of the right size. is this fine as well?

the general approach is that usually, the `key` / `secret` is known at point X in the program, and can be validated and used. at a later point in the program, data to-be-encrypted is present (or to-be-decrypted). usually, this data is present with a nonce --> it makes sense to have a separate key validation from nonce validation.

now, since decrypt already results in an option, we could have both encrypt and decrypt return a result (would make the client code more inconvenient, but safer). opinions? or other ideas?

tests/test_cipher.ml
```diff
@@ -135,34 +135,32 @@ let aes_ctr_cases =
...
-    let { message = pdata ; tag = ptag } =
-      AES.GCM.decrypt ~key ~iv ~adata cdata
+  let case ~key ~p ~a ~nonce ~c ~t =
+    (AES.GCM.of_secret (vx key), vx p, vx a, vx nonce, vx c, vx t) in
+
+  let check (key, p, adata, nonce, c, t) _ =
+    let cipher = AES.GCM.authenticate_encrypt ~key ~nonce ~adata p in
+    let pdata =
+      match AES.GCM.authenticate_decrypt ~key ~nonce ~adata cipher with
+      | None -> assert false
```
dinosaure, line 145: Even if it seems obvious that what we encoded can safely be decoded, we should put a comment to explicit it on the same line - my usual case is to `grep "assert false"`.

hannesm: thanks. I got rid of `assert false` in that file, and report an appropriate test failure now.

hannesm:
@dinosaure 
> module type AEAD = Aead.AEAD

thanks. done.

hannesm:
note: should use constant time equality testing from eqaf in decrypt when comparing the tags/macs (atm: Cstruct.equal)

hannesm:
CI is passing (apart from windows, which opam-repository lags behind).

PR #76, opened 2020-07-14
add ppc64le support

### LINKED ISSUE #75: No known cycle-counting instruction on ppc64
`mirage-crypto` fails to install on ppc64:

```
#=== ERROR while compiling mirage-crypto.0.8.1 ================================#
# context              2.0.7 | linux/ppc64 | ocaml-base-compiler.4.10.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.10/.opam-switch/build/mirage-crypto.0.8.1
# command              ~/.opam/4.10/bin/dune build -p mirage-crypto -j 79
# exit-code            1
# env-file             ~/.opam/log/mirage-crypto-8-9e20af.env
# output-file          ~/.opam/log/mirage-crypto-8-9e20af.out
### output ###
#       ocamlc src/entropy_cpu_stubs.o (exit 2)
# (cd _build/default/src && /home/opam/.opam/4.10/bin/ocamlc.opt -g -I /home/opam/.opam/4.10/lib/bigarray-compat -I /home/opam/.opam/4.10/lib/cstruct -I /home/opam/.opam/4.10/lib/eqaf -I /home/opam/.opam/4.10/lib/eqaf/bigstring -I /home/opam/.opam/4.10/lib/eqaf/cstruct -ccopt -O2 -ccopt -fno-strict-aliasing -ccopt -fwrapv -ccopt -fPIC -ccopt --std=c99 -ccopt -Wall -ccopt -Wextra -ccopt -Wpedantic -ccopt -O3 -ccopt -g -o entropy_cpu_stubs.o native/entropy_cpu_stubs.c)
# native/entropy_cpu_stubs.c: In function 'caml_cycle_counter':
# native/entropy_cpu_stubs.c:86:2: error: #error ("No known cycle-counting instruction.")
#  #error ("No known cycle-counting instruction.")
#   ^~~~~
# native/entropy_cpu_stubs.c:88:1: warning: control reaches end of non-void function [-Wreturn-type]
```

Any suggestions on how to get this working?

avsm:
Implementation of rtdsc based on clang's builtin version and from gperftools at https://chromium.googlesource.com/external/gperftools/+/master/src/base/cycleclock.h

Fixes #75

Just testing this on the ocaml-ci first so that we can have ppc64 CI before undrafting this.

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -82,6 +82,10 @@ CAMLprim value caml_cycle_counter (value __unused(unit)) {
   return Val_long (__rdtsc ());
 #elif defined (__arm__) || defined (__aarch64__)
   return Val_long (read_virtual_count ());
+#elif defined(__powerpc64__)
+    uint64_t rval;
+    __asm__ volatile ("mfspr %0, 268":"=r" (rval));
+    return rval;
 #else
 #error ("No known cycle-counting instruction.")
 #endif
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,5 @@
+* Add support for ppc64le. (@avsm)
+
 ## v0.8.1 (2020-07-02)
 
 * Add Chacha20 implementation (based on abeaumont/ocaml-chacha), supporting
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
-* Add support for ppc64le. (@avsm)
+## v0.8.2 (2020-07-25)
+
+* Add support for ppc64le. (#76 @avsm)
 
 ## v0.8.1 (2020-07-02)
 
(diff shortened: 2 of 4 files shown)


src/native/entropy_cpu_stubs.c
```diff
@@ -82,6 +82,10 @@ CAMLprim value caml_cycle_counter (value __unused(unit)) {
   return Val_long (__rdtsc ());
 #elif defined (__arm__) || defined (__aarch64__)
   return Val_long (read_virtual_count ());
+#elif defined(__powerpc64__) || defined(__ppc64__)
```
hannesm, line 85: do we need to check both `powerpc64` and `ppc64`? if so, this should be done below as well.

what about poly1305-donna.c (which as well has a `#defined(__aarch64__)`)...?

avsm: looks like __powerpc64__ works on the distros we test (gcc), so I'll just use that. We'll get a compile error for other compilers and can fix them as we encounter them.

Also added define check for poly1305

avsm:
Ready for review -- we now have arm32/arm64/ppc64le/i386 CI activated here, so this PR passes (and once merged/released, I can remove the custom pin from the ppc64le CI worker)

PR #77, opened 2020-07-27
Namespace `poly1305_init` to `mcrypto_poly1305_init` in header files

anmonteiro:
Linking mirage-crypto in a project with OpenSSL (and thus, `lcrypto`), resulted in the following error:

```
x86_64-unknown-linux-musl-ld: openssl-1.1.1g-x86_64-unknown-linux-musl/lib/libcrypto.a(poly1305-x86_64.o): in function `poly1305_init':
/build/openssl-1.1.1g/crypto/poly1305/poly1305-x86_64.s:16: multiple definition of `poly1305_init';
  mirage-crypto-0.8.1-x86_64-unknown-linux-musl/libmirage_crypto_stubs.a(poly1305-donna.o):(.text+0x140): first defined here
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
```

### DIFF
--- a/src/native/poly1305-donna-32.h
+++ b/src/native/poly1305-donna-32.h
@@ -44,7 +44,7 @@ U32TO8(unsigned char *p, unsigned long v) {
 }
 
 void
-poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
+mcrypto_poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
 	poly1305_state_internal_t *st = (poly1305_state_internal_t *)ctx;
 
 	/* r &= 0xffffffc0ffffffc0ffffffc0fffffff */
--- a/src/native/poly1305-donna-64.h
+++ b/src/native/poly1305-donna-64.h
@@ -53,7 +53,7 @@ U64TO8(unsigned char *p, unsigned long long v) {
 }
 
 void
-poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
+mcrypto_poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
 	poly1305_state_internal_t *st = (poly1305_state_internal_t *)ctx;
 	unsigned long long t0,t1;
 
--- a/src/native/poly1305-donna.c
+++ b/src/native/poly1305-donna.c
@@ -52,7 +52,7 @@ poly1305_update(poly1305_context *ctx, const unsigned char *m, size_t bytes) {
 
 //stubs for OCaml
 CAMLprim value mc_poly1305_init (value ctx, value key, value off) {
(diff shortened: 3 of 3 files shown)


hannesm:
thanks for the report. is this the only symbol that clashes? which version of openssl do you link against?

I'm not keen on modifying sources we imported from elsewhere. if we really need to, I'd prefer a common prefix (so far we use `mc_` and `_mc_` -- but not `mcrypto_`), and a note at the top of the file (why and what was renamed).

dinosaure:
We probably just need to preprend with `static` keyword:
```diff
diff --git a/src/native/poly1305-donna-32.h b/src/native/poly1305-donna-32.h
index e2af8af..fc9b08a 100644
--- a/src/native/poly1305-donna-32.h
+++ b/src/native/poly1305-donna-32.h
@@ -43,7 +43,7 @@ U32TO8(unsigned char *p, unsigned long v) {
        p[3] = (v >> 24) & 0xff;
 }
 
-void
+static void
 poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
        poly1305_state_internal_t *st = (poly1305_state_internal_t *)ctx;
 
diff --git a/src/native/poly1305-donna-64.h b/src/native/poly1305-donna-64.h
index a9b8f7a..988f3fb 100644
--- a/src/native/poly1305-donna-64.h
+++ b/src/native/poly1305-donna-64.h
@@ -52,7 +52,7 @@ U64TO8(unsigned char *p, unsigned long long v) {
        p[7] = (v >> 56) & 0xff;
 }
 
-void
+static void
 poly1305_init(poly1305_context *ctx, const unsigned char key[32]) {
        poly1305_state_internal_t *st = (poly1305_state_internal_t *)ctx;
        unsigned long long t0,t1;
```

Then, the symbol disappear:
```sh
$  nm libmirage_crypto_stubs.a|grep poly1305_init
00000000000008b0 T mc_poly1305_init
```

hannesm:
@dinosaure 

> add `static` keyword

that sounds good, would you mind PRing that (I guess for all the poly1305 functions, maybe recheck that chacha20 has them in place appropriately as well)

hannesm:
@dinosaure sorry for the noise, I went ahead and did it in c332a15 (part of #78)

hannesm:
@anmonteiro would you mind to test your linking with #78 applied? thanks.

anmonteiro:
Thank you both, @dinosaure and @hannesm. The solution in #78 fixes this for me.

hannesm:
@anmonteiro thanks for confirmation, merged into master and released 0.8.3 with that fix (https://github.com/ocaml/opam-repository/pull/16893)

PR #78, opened 2020-07-27
new test_entropy: testing bootstrap function and timer (cycle counter) to not produce the same output on two subsequent calls

hannesm:
- use `Val_long` in cycle_counter on powerpc64
- mark poly1305 functions static (alternate fix for #77)

### DIFF
--- a/rng/entropy.ml
+++ b/rng/entropy.ml
@@ -29,10 +29,10 @@
 
 module Cpu_native = struct
 
-  external cycles : unit -> int  = "caml_cycle_counter" [@@noalloc]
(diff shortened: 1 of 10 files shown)


hannesm:
NB for fixing #77, I manually ran `readelf -s _build/default/src/libmirage_crypto_stubs.a | grep GLOBAL | grep -v ' mc_' | grep -v ' _mc_' | grep -v ' UND '` to ensure that there aren't any non-`_mc_`  or non-`mc_` prefixed global symbols leaking.

Issue #79, opened 2020-08-10
Bus error in xor_into on arm32


talex5:
Looks like `xor_into` assumes its arguments are 8-byte aligned, but this isn't always the case.
I think this is the cause of the capnp-rpc unit-test failures on arm32 in CI (e.g. https://ci.ocamllabs.io:8100/job/2020-08-10/095735-ci-ocluster-build-9830f6).

The error is:

```
test_lwt alias test-lwt/runtest (got signal BUS)
```

Here's a gdb stacktrace:

```
#0  0x012a6dca in xor_into (n=<optimized out>, 
    dst=0x2ec4b58 "\303C\235\243\020\351\317\311\063\350\003w\224\214\006\337\365\237\263r\223\325ytN\ag\211\063\213Ut\207\325R\274\200\210\064\260OK\255\315\377\332\276\212\327\235\350~\265\060\061\023d\006\260\265\301}\251<\033\344P%Kax\r\335l\373Y\356\231\207\020\313\202p~D\340}\030b&X?\002\306\232y\034\222x\220\024\061F\203*\371uQ\277\341\207\356\342\344\263\303\234r\377\252\340\372}a\016\203\241\326\024@\321F\327_i\006yh\215\234\235\026\001\310o\a_\210\257K\262e2Y-\224\326\034\327\237}avRS2\354(\342\245|^oI\344Z\275\200\277]W\203g\323\306\333~\251m\241\341%Au\233\326\303P<m"..., 
    src=0x2ec1ed1 "\310C\237\f\020\351\315b3\352\245G\026\216\244\357w\236\071\322\220\327xvL\017=\372\366\334\365MLLb\261\206\201\036\066\a\315Z\300\376\333\265\217\327\255\370O\273") at native/misc.c:6
#1  mc_xor_into_generic (b1=<optimized out>, off1=<optimized out>, b2=<optimized out>, off2=<optimized out>, n=n@entry=1385)
    at native/misc.c:51
#2  0x012a6fec in mc_xor_into (b1=<optimized out>, off1=<optimized out>, b2=<optimized out>, off2=<optimized out>, n=1385)
    at native/misc_sse.c:36
#3  0x00e105ca in camlMirage_crypto__Cipher_block__encrypt_843 () at src/cipher_block.ml:223
#4  0x00e10eb0 in camlMirage_crypto__Cipher_block__authenticate_decrypt_984 () at src/cipher_block.ml:300
[...]

PR #81, opened 2020-08-17
test case for #79 (unaligned GCM operation)

hannesm:
not ready for being merged; testing issue report #79


dinosaure:
What about an unroll of the loop if the performance is really important such as:
```c
for (i = arr_size - 4; i >= 0; i -= 4)
{
  dst[i] ^= src[i];
  dst[i+1] ^= src[i+1];
  dst[i+2] ^= src[i+2];
  dst[i+3] ^= src[i+3];
}
```

I suspect that `gcc` is able to optimize such pattern and directly use `uint32_t` instead on some architectures such as `x86_64` and let this code on `arm32`.

hannesm:
@dinosaure it is speculation, I'd first look into benchmarks (on non-virtualized hardware) before committing to any implementation. there's likely some more of these patterns (at least in misc.c & misc_sse.c) -- casting from small integer type to bigger, potentially misaligned.

hannesm:
some more notes:
- passing `-Wcast-align` (in cflags, at least clang ships with this flag) uncovers more issues in the codebase
- a possible fix is to `memcpy` (or `memmove`) the data in question (and I can barely see a performance impact when using memcpy in contrast to the current code)
- bigarray allocation uses malloc, which according to posix uses "properly aligned region"
- the dst / key buffers are usually direct cstructs (allocated by mirage-crypto), and thus properly aligned. what is not aligned may be cstruct's passed in by the user(s)

hannesm:
I pushed 9b30099 which fixes the `xor_into` issue reported in #79.

I also reviewed (with `-Wcast-align`) reported issues:
- aes_aesni (only used on x86_64): key (always properly aligned as outlined above), dst as well, but not src -- since this is only used on x86_64, should be OK
- des_generic: only using key
- ghash*: always use a bigarray (which uses malloc for allocation, and thus is aligned)
- hash_stubs & poly1305: cast from `Bytes_val` to `ctx` - safe since Bytes_val is (aligned!) to the first byte (is this assumption correct?)
- md5/sha1/256/sha512: TODO in _mc_sha1_update (data is from extern, and may be not aligned)
- misc* (counter): all allocated aligned (bigarray or bytes) in cipher_block.ml, no need for memcpy

I'll try to figure out the SHA and MD5 update situation more clearly (by writing some more (regression) tests). I'd be happy for anyone else to look into the reasoning of this comment and above and into the OCaml and C code to get a clearer idea how to move forward.

hannesm:
I don't seem to be able to trigger misalignment issues in the hash implementations.

hannesm:
the two most recent commits modify the entropy gathering and testing, since github actions showed some spurious failures:
the `cpu_rng_bootstrap` used to return 0 if rdseed/rdrand didn't work well (i.e. not returning a valid random after 10 retries), now it raises a `Failure` in this situation. The `bootstrap` function catches such an exception and executes the whirlwind_bootstrap in such a case. Previously the RNG may have been seeded with all-0 (though unlikely) -- now this leads to more "whirlwind non-deterministic execution times" computations at bootstrap.

hannesm:
I moved the RNG changes from here to a separate pull request #82

PR #84, opened 2020-10-13
misc: detect CPU architecture from C compiler

EduardoRFS:
This allows to cross-compile from x86_64 -> ARM64.

### DIFF
--- a/.github/workflows/esy.yml
+++ b/.github/workflows/esy.yml
@@ -1,0 +1,59 @@
+name: Crypto
+
+on: [push, pull_request]
+
+jobs:
+  cross_compile:
+    name: Cross compile
+
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [android.arm64, linux.musl.arm64]
+        ocaml-version: ["~4.10.1000"]
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Create esy wrapper
+        run: |
+          echo '{
+            "name": "cross-compile",
+            "dependencies": {
+              "ocaml": "${{ matrix.ocaml-version }}",
+              "mirage-crypto": "./mirage-crypto.opam",
+              "mirage-crypto-pk": "./mirage-crypto-pk.opam",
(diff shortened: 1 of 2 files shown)


hannesm:
this looks fine to me (travis failure is unrelated).. do you know of a CI setup to test cross-compilation (to ensure we don't break it in the future)?

rizo:
I can confirm that this patch resolves the issue. We have a private cross-compiling CI that targets ARM and uses some mirage packages, would be nice to make it available for open-source projects at some point.

EduardoRFS:
@hannesm My tooling uses esy for that, if you don't mind I can setup a CI to build for Android ARM64

hannesm:
@EduardoRFS if you can add a commit here that extends e.g. the GitHub actions (or travis) with such a job, that'd be great. Thanks a lot.

hannesm:
@EduardoRFS thanks for your second commit, somehow the CI job fails to find ndk.zip - is it expected to work, should I restart the action, or do you have an adjustment to the runner for this?

EduardoRFS:
@hannesm I'm debugging it, it's something related to the shell used on the CI, I should finish it today

hannesm:
@EduardoRFS great, thanks. if you rebase onto master, the travis CI should also turn green.

EduardoRFS:
Ok now it should work

## Dependencies

All dependencies are extracted from opam and every commit `esy solve` will look for a new version, if no dependency changed it will keep cached.

Except, `mtime` has a pinned commit to dune-universe. Because to build to another platform is way easier to just use the dune version

## Missing

Only the `mirage-crypto-pk` is missing, as it depends on `gmp` and I didn't figure it out a way to build it for Android.

EduardoRFS:
@hannesm sorry for all the CI noise. Zarith and GMP was a small problem and there was also an intermittent compile bug(solved)

hannesm:
@EduardoRFS thanks for your work on this, one minor nit: could you squash the last 3 commits (CI-related) into one. then this is fine to be merged. I can also do a release to opam-repository if this eases your workflow.

hannesm (approved):

PR #86, opened 2020-10-17
upgrade to dune-lang 2, remove minios-based xen support, add solo5-based xen support, add unikernel for testing

hannesm:
dune 2 deprecated c_names & c_flags in a stanza, use foreign_library instead


dinosaure:
You should, at least, add a new `.ocamlformat` file with:
```
version = 0.15.0
disable = true
```

dinosaure:
Just a curious question, why you choose to use `Makefile` instead of `cflags.sh`?

hannesm:
@dinosaure based on the discussion in mirage/mirage-xen#23, when using dune's support for compiling C code it always embeds some C flags in the process (`-fPIC`), and also builds both an archive and a shared object -- which is not what we want here.

dinosaure:
> @dinosaure based on the discussion in mirage/mirage-xen#23, when using dune's support for compiling C code it always embeds some C flags in the process (-fPIC), and also builds both an archive and a shared object -- which is not what we want here.

Hmmhmm, that means I need to do an update on `digestif` and `checkseum`, it's correct?

hannesm:
@dinosaure yes, I propose to wait until mirage-xen 6.0.0 is out in the open, and then revise once all our C artifacts. There'll be another round of revising them when either mirage or dune is enhanced with different cross-compilation support.

hannesm:
having said this, at the moment there's no need for action since we're (by luck) getting usable unikernels -- in general mixing PIC and non-PIC compiled objects is undefined behaviour, and a showstopper for PIE binaries.

hannesm:
This is ready for review, CI is green, and awaiting your feedback. I additionally tested it full-circle with a solo5 MirageOS unikernel successfully.

This PR consists of 3 commits: (a) use dune 2.x (in preparation for b, maintenance), (b) use a Makefile for cross-compilation (see mirage/mirage-xen#23) to have control over CFLAGS, (c) remove minios-based xen support, add solo5-based xen support.

PR #87, opened 2020-10-22
ci: fix cross compile

EduardoRFS:
This should fix the cross compile CI and make it stable, as the `generate` tool actually depends on `mirage-crypto` after the update the old patch broke.

This shouldn't happen anymore as `mirage-crypto` now has no patch, but if it starts to fail most of the time is updating it to the latest commit should be enough


hannesm:
thanks @EduardoRFS

hannesm (approved):

PR #89, opened 2020-11-02
freestanding: declare dependencies accurately

hannesm:


freestanding/Makefile
```diff
@@ -12,7 +12,7 @@ else
 CC ?= cc
 FREESTANDING_CFLAGS := $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --cflags ocaml-freestanding)
 MIRAGE_CRYPTO_CFLAGS := $(shell sed 's/^(\(.*\))$$/\1/' ../src/cflags.sexp | tr -d '"')
-CFLAGS := $(MIRAGE_CRYPTO_CFLAGS) $(FREESTANDING_CFLAGS)
+CFLAGS := -I../src/native $(MIRAGE_CRYPTO_CFLAGS) $(FREESTANDING_CFLAGS)
```
dinosaure, line 15: It's really necessary when you copy any files from `native`?

hannesm: according to the cirrus failure https://cirrus-ci.com/task/4742836140638208 it is. i suspect another solution may be to add the .h files to the dependencies, but in the end hacl/fiat/mirage-crypto are now uniform in respect to cross-compiling: they include the `../src/native` and don't bother copying the header files. I just amended this PR to only copy .c files.

PR #90, opened 2020-11-15
mirage-crypto-rng-async

seliopou:
This is an implementation of the async-analogue of `Mirage_crypto_rng_lwt`. It's exposes a single `initialize` function that schedules several periodic jobs to collect entropy from the system and feed it into the `Entropy` module for use by the PRNGs.

I haven't tested this yet, but wanted to throw this up here to gauge whether this would be a welcome contribution. If this looks good, I will follow it up with getting @dinosaure's async ocaml-tls implementation into a state where it can be merged into that project as well.

### DIFF
--- a/mirage-crypto-rng-async.opam
+++ b/mirage-crypto-rng-async.opam
@@ -16,12 +16,12 @@ depends: [
   "ocaml" {>= "4.08.0"}
   "dune" {>= "2.6"}
   "dune-configurator" {>= "2.0.0"}
-  "async" 
+  "async"
   "logs"
   "mirage-crypto"
   "mirage-crypto-rng"
 ]
-conflicts: [ "mirage-runtime" {< "3.8.0"} ]
+available: os != "win32"
 description: """
 
 Mirage-crypto-rng-async feeds the entropy source for Mirage_crypto_rng-based
--- a/mirage-crypto-rng-async.opam
+++ b/mirage-crypto-rng-async.opam
@@ -1,0 +1,29 @@
+opam-version: "2.0"
+homepage:     "https://github.com/mirage/mirage-crypto"
+dev-repo:     "git+https://github.com/mirage/mirage-crypto.git"
+bug-reports:  "https://github.com/mirage/mirage-crypto/issues"
+doc:          "https://mirage.github.io/mirage-crypto/doc"
+authors:      ["David Kaloper <dk505@cam.ac.uk>" "Hannes Mehnert <hannes@mehnert.org>" ]
+maintainer:   "Hannes Mehnert <hannes@mehnert.org>"
+license:      "ISC"
+synopsis:     "Feed the entropy source in an Async-friendly way"
+
+build: [ ["dune" "subst"] {pinned}
+         ["dune" "build" "-p" name "-j" jobs ]
+         ["dune" "runtest" "-p" name "-j" jobs] {with-test} ]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.6"}
+  "dune-configurator" {>= "2.0.0"}
+  "async" 
+  "logs"
+  "mirage-crypto" {=version}
+]
+conflicts: [ "mirage-runtime" {< "3.8.0"} ]
+description: """
+
+Mirage-crypto-rng-async feeds the entropy source for Mirage_crypto_rng-based
+random number genreator implementations, in an Async-friendly way.
+"""
+
--- a/mirage-crypto-rng-async.opam
+++ b/mirage-crypto-rng-async.opam
@@ -19,6 +19,7 @@ depends: [
   "async" 
   "logs"
   "mirage-crypto"
+  "mirage-crypto-rng"
 ]
 conflicts: [ "mirage-runtime" {< "3.8.0"} ]
 description: """
--- a/mirage-crypto-rng-async.opam
+++ b/mirage-crypto-rng-async.opam
@@ -18,7 +18,7 @@ depends: [
   "dune-configurator" {>= "2.0.0"}
   "async" 
   "logs"
-  "mirage-crypto" {=version}
+  "mirage-crypto"
 ]
 conflicts: [ "mirage-runtime" {< "3.8.0"} ]
 description: """
--- a/rng/async/mirage_crypto_rng_async.ml
+++ b/rng/async/mirage_crypto_rng_async.ml
@@ -1,0 +1,88 @@
+open Core
+open Async
+
+open Mirage_crypto_rng
+
+let src = Logs.Src.create "mirage-crypto-rng-async" ~doc:"Mirage crypto RNG Async"
+module Log = (val Logs.src_log src : Logs.LOG)
+
+let running = ref false
+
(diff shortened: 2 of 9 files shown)


hannesm (commented):
Thanks for your PR.
I have not used async (yet) - this looks like a good addition. I'd be happy to merge this PR (apart from the question about blocking behaviour of `getrandom`).

rng/async/mirage_crypto_rng_async.ml
```diff
@@ -0,0 +1,88 @@
...
+
+let periodically_collect_getrandom_entropy time_source span =
+  let source = Entropy.register_source "getrandom" in
+  Synchronous_time_source.run_at_intervals
+    time_source
+    span
+    (fun () ->
+      let per_pool = 8 in
+      let size = per_pool * pools None in
+      (* Might block so run it in a thread. *)
```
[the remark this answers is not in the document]
seliopou: Ack. (Side note: `In_thread.run` uses a thread pool under the hood.) The periodic call doesn't need to handle this case, since if it hasn't been initialized the call to `getrandom_init` in `initialize` will block. Speaking of which, what about that call?

mirage-crypto-rng-async.opam
```diff
@@ -0,0 +1,29 @@
...
+         ["dune" "build" "-p" name "-j" jobs ]
+         ["dune" "runtest" "-p" name "-j" jobs] {with-test} ]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.6"}
+  "dune-configurator" {>= "2.0.0"}
+  "async" 
+  "logs"
+  "mirage-crypto"
```
seliopou, line 21: I had to remove the version constraint so i could pin install this locally. Also I suspect the CI will pass now, but I don't think this is what should be merged to master. Any suggestions.

rng/async/mirage_crypto_rng_async.ml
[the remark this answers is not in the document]
hannesm: > Speaking of which, what about that call in `initialize`?

As far as I read (Linnux glibc manpage, FreeBSD source code), the `getandom` blocking behaviour is only there during bootup -- on FreeBSD once init is called, the RNG is initialized and will not ever block. From my point of view, this is sane behaviour - OCaml programs using Mirage_crypto_rng are blocking in case the host system RNG is not ready. Do you prefer a different semantics (or have experienced in scenarios that getrandom/getentropy is blocking and this is undesired behaviour)?

Since we're mixing RDRAND & getrandom & timer (rdtsc), we could as well make the getrandom optional (i.e. there's likely sufficiently good random with RDRAND & timer), but taking into account I never observed the blocking state of getrandom, I'd leave it as is. As mentioned, if you're observing the blocking `getrandom`, let us discuss this again.

mirage-crypto-rng-async.opam
hannesm: that's fine for now - I'll remember to include the `{= version}` constraint at release-time.

rng/async/mirage_crypto_rng_async.ml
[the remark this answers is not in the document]
seliopou: Sounds like it's ok to leave things the way they are.

seliopou:
So a few things to deal with: the win32 build is broken and, reading `.github/test.yml`, I'm not sure exactly why. Second, some testing. I created [this gist](https://gist.github.com/seliopou/67ca2a75a12e78962fb6f30a9dcc4c24) just to verify that initialization was unique across runs and timers were being scheduled properly.

dinosaure:
`async` does not support Windows, you should may be update this line:
https://github.com/mirage/mirage-crypto/blob/de521205fe4f14178f8e07ebce8063e355303d69/.github/workflows/test.yml#L33

hannesm:
@seliopou 
> So a few things to deal with: the win32 build is broken and, reading `.github/test.yml`, I'm not sure exactly why.

Took me a while to figure out what the issue is. Turns out, the windows runner uses a different opam-repository (https://github.com/fdopen/opam-repository-mingw/) - where async is marked as unavailable on windows. To have the CI succeed, I guess all that is needed is to specify `available: os != "win32"` in mirage-crypto-rng-async.

> Second, some testing. I created this gist just to verify that initialization was unique across runs and timers were being scheduled properly.

That looks good. There is `test_entropy.ml` which checks that individual entropy sources do not output identical data. There's as well `test_entropy_collection` which "tests" the mirage-crypto-rng-mirage implementation (there's not really a test unfortunately).

For lwt (and async), it would be good to have tests which validate "that initialization was unique across runs and timers were being scheduled properly" (but I haven't bothered developing such a test so far). Your gist looks good, it may be worth to put it in the `test` subdirectory.

rng/async/mirage_crypto_rng_async.ml
```diff
@@ -0,0 +1,81 @@
...
+let ns_since_epoch time_source () =
+  Synchronous_time_source.now time_source
+  |> Time_ns.to_int_ns_since_epoch
+  |> Int64.of_int
+
+let periodically_collect_cpu_entropy time_source span =
+  Synchronous_time_source.run_at_intervals
+    time_source
+    span
+    (fun () -> Entropy.cpu_rng None)
```
seliopou, line 20: On a RaspberryPI 2, this doesn't collect any entropy. Somewhat orthogonal to this PR, but it might be nice to expose something like `val Entropy.cpu_native_support : bool` so that implementations can check if it's worthwhile creating the timer.

seliopou:
I updated the interface to make it easier to write the entropy collection test. Assuming CI passes, I think this is good to go?

Though, see my note about cpu native entropy sources inline.

rng/async/mirage_crypto_rng_async.ml
```diff
@@ -0,0 +1,81 @@
...
+  let data = Mirage_crypto_rng_unix.getrandom 128 in
+  Entropy.header i data
+
+let initialize ?g ?time_source ?(sleep = Time_ns.Span.of_int_sec 1) generator =
+  let time_source =
+    Option.value ~default:(Synchronous_time_source.wall_clock ()) time_source
+  in
+  if !running then
+    Log.debug
+      (fun m -> m "Mirage_crypto_rng_lwt.initialize has already been called, \
```
hannesm, line 52:
```suggestion
      (fun m -> m "Mirage_crypto_rng_async.initialize has already been called, \
```

rng/async/mirage_crypto_rng_async.ml
```diff
@@ -0,0 +1,81 @@
...
+                   ignoring this call.")
+  else begin
+    (try
+       let _ = default_generator () in
+       Log.warn (fun m -> m "Mirage_crypto_rng.default_generator has already \
+                             been set (but not via \
+                             Mirage_crypto_rng_lwt.initialize). Please check \
```
hannesm, line 59:
```suggestion
                             Mirage_crypto_rng_async.initialize). Please check \
```

hannesm (approved):

hannesm:
thanks spiros, I squash-merged this to master :)

PR #92, opened 2020-12-05
require client code to match on cpu_rng support

seliopou:
To avoid scheduling timers that do no work, require client code to match on the presence of `cpu_rng` support before attempting to use it.

It appears that `cpu_rng_bootstrap` was kindof already doing this, but with an exception. In this cases, this PR just makes what's going on a little more explicit.


seliopou:
It's a lot easier to evaluate the diff by looking at `git show --patience --word-diff=color`.

hannesm:
thanks spiros -- rebased and re-added some exception handling to bootstrap_cpu_rng.

PR #95, opened 2020-12-29
mirage-crypto-rng: revise Entropy.cpu_rng to delay entropy feeding.

### LINKED ISSUE #94: Memory leak in RNG _sources?
According to `perf`, ocluster-scheduler is spending a lot of time in `List.length`:
```
-   57.92%  ocluster-schedu  ocluster-scheduler  [.] camlStdlib__list__length_aux_83
     camlStdlib__list__length_aux_83
     camlMirage_crypto_rng__Entropy__register_source_143
     camlMirage_crypto_rng__Entropy__cpu_rng_596
     camlMirage_crypto_rng_lwt__one_151
     camlLwt__callback_1288
     camlLwt__iter_callback_list_873
     camlLwt__run_in_resolution_loop_926
     camlLwt__resolve_946
     camlLwt__wakeup_general_983
     caml_start_program
```

I don't really understand this code, but it looks like `initialize` calls `rdrand_task` https://github.com/mirage/mirage-crypto/blob/ab6c5f323d7a848a85ecaa091c380076f3c11d1a/rng/lwt/mirage_crypto_rng_lwt.ml#L61

That then calls [`Entropy.cpu_rng`](https://github.com/mirage/mirage-crypto/blob/ab6c5f323d7a848a85ecaa091c380076f3c11d1a/rng/entropy.ml#L139) once per second. Each time, that calls `register_source`, which gets the length of `_sources` and then appends to it. As far as I can see, nothing is ever removed from this list.

hannesm:
This fixes #94, since register_source is called only once for the rdrand task.

### DIFF
--- a/rng/async/mirage_crypto_rng_async.ml
+++ b/rng/async/mirage_crypto_rng_async.ml
@@ -17,7 +17,7 @@ let periodically_collect_cpu_entropy time_source span =
   Synchronous_time_source.run_at_intervals
     time_source
     span
-    (fun () -> Entropy.cpu_rng None)
+    (Entropy.cpu_rng None)
 
 let periodically_collect_getrandom_entropy time_source span =
   let source = Entropy.register_source "getrandom" in
--- a/rng/entropy.ml
+++ b/rng/entropy.ml
@@ -116,7 +116,8 @@ let bootstrap id =
   try cpu_rng_bootstrap id with Failure _ -> whirlwind_bootstrap id
 
 let interrupt_hook () =
-  let buf = Cstruct.create 4 in fun () ->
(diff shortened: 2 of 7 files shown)


hannesm:
@EduardoRFS as I don't understand esy too well, could you briefly look into the cross-compilation CI failure? to me it looks that it tries to build with a released / master version, and fails since the interface changed...

EduardoRFS:
@hannesm feel free to merge, I will fix it in a couple of hours, this CI style isn't great

EduardoRFS:
The CI should be green for the next run, this was fixed by https://github.com/esy-ocaml/esy-opam-override/pull/86

PR #99, opened 2021-01-20
Fix computation of d = e^-1 mod lam n, with lam = lcm (p - 1) (q - 1)

hannesm:
Spotted by @psafont

This is an alternate fix for #62, and adheres to RFC 8017 (etc.), NIST.


psafont (commented):
Could we add a unit-test derived from the key in https://github.com/mirage/mirage-crypto/issues/62 to validate this?

hannesm:
@psafont sure, added in 17eaa2d

psafont (approved):

PR #100, opened 2021-01-21
pk/rsa: use correct invariant when checking parameter d

psafont:
The correct invariant is `d ≡ e⁻¹ (mod λ(n))`, this means that the modulus is applied to both `d` and `e⁻¹`, which was not done with the previous check.

In practice this means that RSA keys using `d = (p - 1) * (q - 1)` instead of `d = lcm (p - 1) (q - 1)` could fail this test when they are valid.

Note: ([FIPS 186-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf#page=62) is more strict and would use the check previously used, but I think it should be an optional check)


hannesm:
thanks

PR #103, opened 2021-03-03
mirage-crypto: define bits if not defined (to support old cpuid.h)

hannesm:


hannesm:
not really worth it - such systems with such old compilers do not have a `getentropy` system call. it is more burden than worth it.

PR #107, opened 2021-03-10
mirage-crypto-ec: add 25519

hannesm:
- [x] add ED25519
- [x] figure out performance (whether to move scalar_mult + inversion to C to avoid lots of OCaml -> C calls)


hannesm:
I evaluated the performance using the 5 existing X25519 implementations (I did not use `hacl` since it is abandoned (superseeded by `hacl-star` and conflicts with `hacl_x25519`) available on opam:
* callipyge: 44.676 ops per second (447 iters in 10.005)
* hacl_x25519 1792.651 ops per second (17774 iters in 9.915)
* rfc7748 200.734 ops per second (2006 iters in 9.993)
* hacl-star 1803.398 ops per second (17882 iters in 9.916)
* mirage-crypto-ec with inversion and scalar multiplication in OCaml (cfe466f)  560.714 ops per second (6350 iters in 11.325)
* mirage-crypto-ec with inversion and scalar multiplication in C (7cc310b) 1714.683 ops per second (17082 iters in 9.962)

Since the C code is boring (taken from BoringSSL), I do not see a reason to retain the arithmetic instructions in OCaml (esp. since it is a pretty high price to pay for all the OCaml<->C calling). Heap allocation still happens on the OCaml side, though memmove and memset is used on the C side.

dinosaure:
Seems very good :tada:!

hannesm:
FWIW, see https://github.com/mirage/mirage-crypto/pull/202#issuecomment-1963060125 for an updated table, and a utility (bench/ec.exe) to generate such a table.

PR #114, opened 2021-03-30
Makefiles: opam 2.1 invocation, make work w/o opam

sternenseemann:


hannesm:
thanks, this is the same as https://github.com/mirage/mirage-crypto/pull/115/commits/a9fba96a9d400a9d83a583ecf6c6cee54d56d969

PR #116, opened 2021-04-04
fix CI esy cross compile and add macOS ARM64 as target

EduardoRFS:
## Why the CI was broken?

- esy@0.6.8 had a bug caused on invalid symlink, that was rarely triggered outside of cross compilation, with `esy@0.6.10` that was fixed
- because `reason-mobile` itself depends on mirage-crypto this CI should be using resolutions and it was not

## Changes

- added mirage-crypto-ec on the cross compile list
- added a cross OS case, `macOS -> Android ARM64`
- added `macos.arm64` as a target, this was also checked locally from `Linux -> macOS ARM64` but this cannot be supported on a CI

### DIFF
--- a/.github/workflows/esy.yml
+++ b/.github/workflows/esy.yml
@@ -9,10 +9,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [android.arm64, linux.musl.arm64]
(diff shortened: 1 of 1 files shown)


dinosaure (approved):
LGTM

hannesm:
thanks

PR #121, opened 2021-04-23
Make inversion run in constant time.

dfaranha:
Hi,

I've worked together with Benjamin Hvass (@bshvass) and Bas Spitters (@spitters) in the Fiat-Crypto implementation of field inversion. We are pretty happy to see someone making use of the code!

Our C templates had a minor timing issue at the end of algorithm that was later fixed, so I'm pushing a manual fix here as well.

PS: I could not run local tests due to lack of a proper build environment, but it should be correct.

Thank you!


hannesm:
Thanks for your PR -- it would be great if you could as well update the upstream (https://github.com/mit-plv/fiat-crypto/tree/master/inversion-c) with the patch (that looks fine). I'll adapt and push so that the `npYYY_AA.h` include the selectznz function, then CI will likely succeed :)

dfaranha:
Thanks! We will push the fix to upstream shortly.

hannesm:
Thanks. The cross-compilation CI seems to use an old commit (@EduardoRFS any insights?). Since other cross-CI work, and all other CI also succeed, I'll merge and release.

PR #122, opened 2021-05-07
Fix C stubs dependency specification in mirage-crypto-ec

TheLortex:
When using `mirage-crypto-ec` with Mirage 4, I've encountered build failures as `mirage_crypto.h` couldn't be found when building `mirage-crypto-ec`'s stubs. Indeed `dune` does not deduce from the `-I../src/native` flag that it has to track `../src/native` as a dependency folder. 

Instead the `include_dirs` option should be used: that's what this PR is about. 

In the opam world this hasn't been a problem because the dune file in `ec-freestanding` correctly specifies the dependency. 

For more information on the dune options for C compilation: https://dune.readthedocs.io/en/stable/concepts.html#foreign-stubs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## unreleased
+
+- mirage-crypto-ec: dune C stubs compilation rules: explicitely declare the 
+  include directory instead of listing it as a flag, so that the dependency
+  is correctly tracked (#122 by @TheLortex)
+
 ## v0.10.1 (2021-04-25)
 
 - mirage-crypto-ec: make inversion run in constant time (#121 by @dfaranha)
--- a/ec/dune
+++ b/ec/dune
@@ -6,8 +6,9 @@
   (language c)
   (names p224_stubs np224_stubs p256_stubs np256_stubs p384_stubs np384_stubs
     p521_stubs np521_stubs curve25519_stubs)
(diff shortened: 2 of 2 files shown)


dinosaure (approved):

hannesm:
the change looks fine to me, though I don't like the wording "fix" (neither in the commit message nor in the changes entry). from my perspective, it wasn't broken before - and if `dune` does not support `-I ..` (or gets confused by its usage), it should complain about its usage and recommend to use `include_dirs`.

if you could rephrase that (and mention _why_ and _what_ you changed in this commit), I'm happy to merge this.

TheLortex:
Thank you for your review. Indeed nothing was "broken" before, but this is by luck, as the header files were grabbed when coincidentally building `ec-freestanding`. To me, we're "fixing" an issue that's upcoming (when the `ec-freestanding` build gets removed) -- but I've changed the changes entry and the commit message.

With regards to dune not supporting `-I ..`, it's just that it doesn't parse C flags, it merely transfers them to the underlying C compiler, that's why the `include_dirs` field exists, and because it's not parsing the flags it cannot complain about the usage of the `-I` option. Additionally, the `extra_deps` field is available to track any dependency that would occur in the C build.

I hope that makes sense to you !

hannesm:
thanks.

PR #129, opened 2021-07-13
Add s390x enablement to mirage-crypto

edelsohn:
The two commits correct the compile failures of mirage-crypto on the IBMz s390x architecture.

Note that the STCK instruction used in the cycle count implementation guarantees a globally unique result, which is expensive. If that is not needed, STCKF instruction is less expensive.

Also s390x has the PRNO instruction to provide a pseudo-random number, if that would be useful in the other prng functions of the cpu entropy stubs file.

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -108,6 +108,16 @@ static inline uint64_t rdcycle64(void)
 }
 #endif
 
+#if defined (__s390x__)
+static inline uint64_t getticks(void)
+{
+  uint64_t rval;
+  __asm__ __volatile__ ("stck %0" : "=Q" (rval) : : "cc");
+  return rval;
+}
+#endif
+
+
 CAMLprim value mc_cycle_counter (value __unused(unit)) {
 #if defined (__i386__) || defined (__x86_64__)
   return Val_long (__rdtsc ());
@@ -117,6 +127,8 @@ CAMLprim value mc_cycle_counter (value __unused(unit)) {
   return Val_long (read_cycle_counter ());
 #elif defined(__riscv) && (64 == __riscv_xlen)
   return Val_long (rdcycle64 ());
+#elif defined (__s390x__)
+  return Val_long (getticks ());
 #else
 #error ("No known cycle-counting instruction.")
 #endif
--- a/src/native/mirage_crypto.h
+++ b/src/native/mirage_crypto.h
@@ -47,7 +47,7 @@ extern struct _mc_cpu_features mc_detected_cpu_features;
 
 #endif /* __mc_ACCELERATE__ */
 
-#if defined (__x86_64__) || defined (__aarch64__) || defined (__powerpc64__) || (64 == __riscv_xlen)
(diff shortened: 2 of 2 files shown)


hannesm:
Thanks. The cycle_count does not need to be globally unique -- it is executed on every event loop enter and the lower 4 bytes are mixed into the entropy pool. It would be great to have some instruction that returns an unpredictable value (at least for a remote attacker) in the lower bytes. If STCKF provides this, that should be fine.

And indeed if there is a similar instruction to RDRAND (on Intel / AMD64 CPUs) that output a random number, that could be added as well (though this code is executed depending on feature detection (cpuid()). Do all s390x CPU provide such an instruction (that would be nice and straightforward to retrieve and feed into the entropy pool).

edelsohn:
STCKF (Store Clock Fast) and PRNO (Perform Random Number Operation) are available in z14 processors and above.  The current generation is z15, and IBM is not focused on support for earlier processors, so there is no concern about cpu feature detection.

IBMz has a number of cryptographic hardware accelerators for operations like Poly1305 and ChaCha20, and the PRNO True Random Number Generator.

edelsohn:
The IBMz PRNO TRNG instruction will return a buffer of random numbers.

`
/**
--
  | * cpacf_trng() - executes the TRNG subfunction of the PRNO instruction
  | * @ucbuf: buffer for unconditioned data
  | * @ucbuf_len: amount of unconditioned data to fetch in bytes
  | * @cbuf: buffer for conditioned data
  | * @cbuf_len: amount of conditioned data to fetch in bytes
  | */
  | static inline void cpacf_trng(u8 *ucbuf, unsigned long ucbuf_len,
  | u8 *cbuf, unsigned long cbuf_len)
  | {
  | register unsigned long r0 asm("0") = (unsigned long) CPACF_PRNO_TRNG;
  | register unsigned long r2 asm("2") = (unsigned long) ucbuf;
  | register unsigned long r3 asm("3") = (unsigned long) ucbuf_len;
  | register unsigned long r4 asm("4") = (unsigned long) cbuf;
  | register unsigned long r5 asm("5") = (unsigned long) cbuf_len;
  |  
  | asm volatile (
  | "0:	.insn	rre,%[opc] << 16,%[ucbuf],%[cbuf]\n"
  | "	brc	1,0b\n"	  /* handle partial completion */
  | : [ucbuf] "+a" (r2), [ucbuflen] "+d" (r3),
  | [cbuf] "+a" (r4), [cbuflen] "+d" (r5)
  | : [fc] "d" (r0), [opc] "i" (CPACF_PRNO)
  | : "cc", "memory");
  | }

`

edelsohn:
Any update on the pull request? Also, any interest from Robur to utilize the IBMz crypto hardware accelerator instructions?

hannesm:
Dear @edelsohn,

thank you for your PR. I think this is fine to go as-is. Further improvements you mention: STCKF, PRNO, POLY1305/CHACHA20 instructions would be nice to have, but we at robur are at the moment busy with other projects (and don't have customer demand for improved s390x) support.

If you have (customer) demand for better random numbers (using PRNO) or improved ciphers, please don't hesitate to contact us.

avsm:
s390x support is now active in ocaml-ci, so this will be tested on trunk shortly...

PR #133, opened 2021-07-23
multicore requires c11 (using the _Atomic keyword in typedefs)

hannesm:
see https://github.com/ocaml-multicore/ocaml-multicore/issues/574

please confirm that this will be what OCaml-multicore will need and use.

//cc @Engil @kayceesrk @avsm @Sudha247

### DIFF
--- a/config/cfg.ml
+++ b/config/cfg.ml
@@ -1,4 +1,4 @@
-let std_flags = ["--std=c99"; "-Wall"; "-Wextra"; "-Wpedantic"; "-O3"]
+let std_flags = ["--std=c11"; "-Wall"; "-Wextra"; "-Wpedantic"; "-O3"]
 
 let () =
   let c = Configurator.V1.create "mirage-crypto" in


Engil:
Hello!

As it currently stands, Multicore is relying on `gnu11` rather than `c11` because of the `asm` keyword usage. (and some others GNU extensions.)
I'm unsure if the end goal is to support strict `c11`, but in the meantime, `gnu11` is the right option.

A more optimal way of handling this would be to extract the correct C flags from the OCaml installation by using `ocamlc -config`. (this way the flags are always as correct as the OCaml installation wants it to be)

cc @ctk21 and @sadiqj as well who may have a better answer.
(and @kit-ty-kate as well with whom we investigated such issues recently)

hannesm:
Thanks for your quick reply. CI-wise it seems the use of the OCaml headers in mirage-crypto are satisfied by C11 (i.e. no need for GNU extensions). I don't see a good reason to use `--std=g11` here in mirage-crypto. Extracting the "C compiler flags" from "ocamlc -config" may not work as intended for multiple reasons:
- optimization level could be decided on a library level (i.e. OCaml runtime may use -O0, a library -O3)
- how CPU features are detected & used (mirage-crypto compiles code e.g. with AESNI instructions on any x86 machine, and decides at runtime which AES implementation to use)
- warnings: in mirage-crypto we appreciate all warnings (-Wall -Wpedantic), and in development mode (as run by CI or by a local dune build) -Werror is passed as well
- C standard: mirage-crypto is mostly C99 (apart from the change in caml/misc.h, which now requires C11), and not "whatever the OCaml runtime used to be compiled with.

Engil:
Thanks for your answer.

It appears indeed using the C flags as returned by `ocamlc -version` is not a very good solution, my apologies for the oversight.

It seems that `gnu11` is not required for mirage-crypto indeed, and `c11` just works.

I just tested your branch against a recent `4.12.0+domains` branch  and it does seems to compile well. (with the assistance of some patches, ocaml-multicore/ocaml-multicore#623)
(and with the testsuite passing, which makes me happy.)

Breaking changes from includes are not entirely dealt with, [other includes](https://github.com/ocaml-multicore/ocaml-multicore/issues/602) may lead to other issues, there's definitely more ironing to be done on our side, and we will communicate on breaking changes when the final interface has been ironed out.

hannesm:
great, thanks @Engil.

PR #134, opened 2021-07-28
use Cstruct.length instead of deprecated Cstruct.len

hannesm:
now requires cstruct >= 6.0.0

### DIFF
--- a/bench/speed.ml
+++ b/bench/speed.ml
@@ -304,7 +304,7 @@ let benchmarks = [
     let open Mirage_crypto_rng.Fortuna in
     let g = create () in
     reseed ~g (Cstruct.of_string "abcd") ;
-    throughput name (fun cs -> generate ~g (Cstruct.len cs))) ;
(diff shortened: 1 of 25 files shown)


hannesm:
I've no clue why the macos jobs are failing now:
```
# Run eval $(opam env --switch=4.12.0) to update the current shell environment
/usr/local/bin/opam install -y depext
Error:  depext unmet availability conditions, e.g. 'opam-version >= "2.0.0~beta5" & opam-version < "2.1"'
Error: Error: The process '/usr/local/bin/opam' failed with exit code 5
```

Any hints @avsm @smorimoto ?

avsm:
Homebrew has upgraded to opam 2.1.0, so the depext invocation can just be removed now I think.

smorimoto:
It's a fair opinion, but in opam 2.1, if depext is installed as a package, rather than a built-in depext, it simply acts like 2.0, so no error should occur.

PR #135, opened 2021-07-28
gh actions: update to opam 2.1 invocation and refresh compiler versions

avsm:


.github/workflows/test.yml
```diff
@@ -19,20 +19,12 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Use OCaml ${{ matrix.ocaml-version }}
-        uses: avsm/setup-ocaml@v1
+        uses: ocaml/setup-ocaml@v2
         with:
           ocaml-version: ${{ matrix.ocaml-version }}
```
smorimoto, line 24:
```suggestion
          ocaml-compiler: ${{ matrix.ocaml-version }}
```

.github/workflows/windows.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Use OCaml ${{ matrix.ocaml-version }}
-        uses: avsm/setup-ocaml@v1
+        uses: ocaml/setup-ocaml@v2
         with:
           ocaml-version: ${{ matrix.ocaml-version }}
```
smorimoto, line 24:
```suggestion
          ocaml-compiler: ${{ matrix.ocaml-version }}
```

smorimoto:
We may have to avoid depext for this only on the Windows runners. (or we can make more fundamental improvements!)
https://github.com/mirage/mirage-crypto/blob/87c1f5bb3539018187f8cc4486faef4a2c74c130/mirage-crypto-rng-async.opam#L24

hannesm:
Thanks for this PR. Unfortunately this changes the set of CI failures, but does not fix all. Namely, the windows CI (that uses a separate configuration file) avoided the installation of mirage-crypto-rng-async -- it looks like the new "setup-ocaml" doesn't care about that.

smorimoto:
2.1 has been temporarily disabled. v2 uses the latest opam < 2.1. Native Windows support is planned to be introduced in 2.2, so we will unlock this constraint when it is released.

smorimoto:
For issues on Windows, it may be enough to simply disable auto depext.

hannesm:
I looked into the ocaml/setup-ocaml documentation (and examples), and tried with a subsequent commit (efb13ff) -- happy to see whether this works as expected (and hear your thoughts). indeed, the mirage-crypto-rng-async opam package is not supported on windows.

PR #136, opened 2021-08-03
Set license to ISC in mirage-crypto-ec.opam

pukkamustard:
Assuming that all packages are licensed under ISC and that this is a typo.

### DIFF
--- a/mirage-crypto-ec.opam
+++ b/mirage-crypto-ec.opam
@@ -20,7 +20,7 @@ authors: [
   "Massachusetts Institute of Technology"
   "Zoe Paraskevopoulou <zoe.paraskevopoulou@gmail.com>"
 ]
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks. It would be nice if this would be the case, but please take a look into `ec/LICENSE.md` I extended it in e8e47454487751071fcc835ac1300f2f2331b471 with the remaining information (BoringSSL, fiat-crypto, code developed only for this library).

TL;DR: mirage-crypo-ec is MIT licensed, and mirage-crypto-rng-mirage is BSD 2-clause licensed. The remaining of this repository is ISC licensed.

hannesm:
And while at it, anyone aware how to teach opam/dune which LICENSE file to install for a given opam package?

PR #139, opened 2021-10-20
drop rresult dependency

hannesm:


hannesm:
thanks @smorimoto

reynir:
Oops, I missed that this was already merged...

PR #142, opened 2021-10-25
p-{256/384/521} public key compression and decompression

Zett98:
NIST P-{256/384/521} public key compression and decompression support.

### DIFF
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -56,10 +56,19 @@ module type Dsa = sig
     val generate : key:priv -> Cstruct.t -> Cstruct.t
   end
 end
+module type Dsa_with_compression = sig
+  include Dsa
+  val pub_to_compressed : pub -> Cstruct.t
+end
 
 module type Dh_dsa = sig
   module Dh : Dh
   module Dsa : Dsa
+end
+
+module type Dh_dsa_with_compression = sig
+  module Dh : Dh
+  module Dsa : Dsa_with_compression
 end
 
 module type Parameters = sig
@@ -172,7 +181,7 @@ module type Point = sig
 
   val of_cstruct : Cstruct.t -> (point, error) result
 
-  val to_cstruct : point -> Cstruct.t
+  val to_cstruct : ?compressed:bool -> point -> Cstruct.t
 
   val to_affine_raw : point -> (field_element * field_element) option
 
@@ -276,7 +285,7 @@ module Make_point (P : Parameters) (F : Foreign) : Point = struct
       Fe.to_bytes out_y y;
       Some (out_x, out_y)
 
-  let to_cstruct p =
+  let[@warning "-27"] to_cstruct ?(compressed=false) p =
     match to_affine p with
     | None -> Cstruct.create 1
     | Some (x, y) ->
@@ -407,7 +416,7 @@ module type Foreign_n = sig
   val to_montgomery : field_element -> field_element -> unit
 end
 
-module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H : Mirage_crypto.Hash.S) : Dsa = struct
+module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H : Mirage_crypto.Hash.S) = struct
   let create () = Cstruct.to_bigarray (Cstruct.create Param.fe_length)
 
   type priv = scalar
@@ -468,9 +477,9 @@ module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H
 
   type pub = point
 
-  let pub_of_cstruct = P.of_cstruct
+  let pub_of_cstruct = P.of_cstruct 
 
-  let pub_to_cstruct = P.to_cstruct
+  let pub_to_cstruct pk = P.to_cstruct pk
 
   let generate ?g () =
     (* FIPS 186-4 B 4.2 *)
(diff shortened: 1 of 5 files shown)


Zett98:
I pushed all the changes, Do I need to change anything or everything is okay?

hannesm:
Thanks for your PR. I'd like to review it in detail when I find time for it. It adds quite a bit of code (that needs to be reviewed in more detail, since it will then need to be maintained within mirage-crypto), and extends the public interface.

ec/mirage_crypto_ec.ml
```diff
@@ -276,7 +285,7 @@ module Make_point (P : Parameters) (F : Foreign) : Point = struct
       Fe.to_bytes out_y y;
       Some (out_x, out_y)
 
-  let to_cstruct p =
+  let[@warning "-27"] to_cstruct ?(compressed=false) p =
```
hannesm, line 288: why is this `warning "-27"` needed?

ec/mirage_crypto_ec.ml
```diff
@@ -584,6 +593,112 @@ module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H
     with
     | Message_too_long -> false
 end
+module type Prime_identity = sig
+  val pident: Cstruct.t
```
hannesm, line 597: this should be added to `Parameters` instead.

ec/mirage_crypto_ec.ml
```diff
@@ -584,6 +593,112 @@ module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H
...
+    let byte_num = i / 8 in
+    let bit_num = i mod 8 in
+    let byte = Cstruct.get_uint8 buf byte_num in
+    byte land (1 lsl bit_num) <> 0
+    
+  let pow =
+    let mult a b = 
+      let r = Fe.create () in 
+      Fe.mul r a b;
+      r in
```
hannesm, line 623: please push the `in` into the next line:
```suggestion
      r
    in
```

ec/mirage_crypto_ec.ml
```diff
@@ -584,6 +593,112 @@ module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H
...
+    let sqr x = 
+      let r = Fe.create () in
+      Fe.sqr r x;
+      r in
```
hannesm, line 627: same as above

ec/mirage_crypto_ec.ml
```diff
@@ -584,6 +593,112 @@ module Make_dsa (Param : Parameters) (F : Foreign_n) (P : Point) (S : Scalar) (H
...
+    let y_struct2 = Cstruct.rev y_struct2 in
+    let ident = Cstruct.get_uint8 pk_cstruct 0 in
+    let signY = 2 + (Cstruct.get_uint8 y_struct (P.byte_length - 2))  land 1 in
+    let res = 
+      if Int.equal signY ident then y_struct else y_struct2 in
+    let out = Cstruct.create ((P.byte_length * 2) + 1) in 
+    Cstruct.set_uint8 out 0 4;
+    Cstruct.blit pk_cstruct 1 out 1 P.byte_length;
+    Cstruct.blit res 0 out (P.byte_length + 1) P.byte_length;
+    out
```
hannesm, line 684: please insert a newline before the next binding `let of_cstruct`

hannesm:
I added some review comments, the main issue I have with this PR is:
> because there's no neede primitives to implement it.

What is `neede primitives`? What would be needed to get this to work with P-224 as well? (That would simplify the modules enormously, and your compress & decompress algorithm looks very generic.)

Zett98:
> I added some review comments, the main issue I have with this PR is:
> 
> > because there's no neede primitives to implement it.
> 
> What is `neede primitives`? What would be needed to get this to work with P-224 as well? (That would simplify the modules enormously, and your compress & decompress algorithm looks very generic.)

my bad with the typo, p-224 requires modular square root(the primitive operation I was referring to) as Params.p mod 4 != 3 mod 4, but there is no square root in the exposed C code, and trying to pull it out of boringssl would require copying a lot of code from there, basically recreating a bignum library here

ec/mirage_crypto_ec.ml
Zett98: compressed variable is unused in the body of this function, added it here so that Point and Point_with_compression would have the same signature

ec/mirage_crypto_ec.ml
Zett98: changed this

ec/mirage_crypto_ec.ml
Zett98: done.

ec/mirage_crypto_ec.ml
Zett98: done.

ec/mirage_crypto_ec.ml
Zett98: done.

hannesm:
Thanks again for your work. I reviewed in a code editor, and reshuffled your work to have fewer changes on the Mirage_crypto_ec interface. Please have a look at https://github.com/hannesm/mirage-crypto/commit/bba579ca2ce4f02eb82dacf8c3723c6202d9bd79 whether this would fit your needs. Also, in case you have an opinion on the following questions, I'd be interested:

On a further note, Dh now accepts compressed points as well. Questions:
- are these API changes fine?
- should Dh.secret_of_cs and Dh.gen_key have an optional ?compress argument?

Zett98:
> Thanks again for your work. I reviewed in a code editor, and reshuffled your work to have fewer changes on the Mirage_crypto_ec interface. Please have a look at [hannesm@62f5b78](https://github.com/hannesm/mirage-crypto/commit/62f5b7822d41d3d50d84fe0732b105cead4b236c) whether this would fit your needs. Also, in case you have an opinion on the following questions, I'd be interested:
> 
> On a further note, Dh now accepts compressed points as well. Questions:
> 
> * are these API changes fine?
> * should Dh.secret_of_cs and Dh.gen_key have an optional ?compress argument?

Thanks for reviewing this, 
1. The api changes are fine
2. Not sure as that is allowed to decompress secrets as they are not compressed.(please correct me if i'm wrong). gen_key def should have an optional compress argument.

hannesm:
> Not sure as that is allowed to decompress secrets as they are not compressed.(please correct me if i'm wrong). gen_key def should have an optional compress argument.

You are right. The DH API is:
```OCaml
  type secret
  val secret_of_cs : Cstruct.t -> (secret * Cstruct.t, error) result
  val gen_key : ?g:Mirage_crypto_rng.g -> unit -> secret * Cstruct.t
  val key_exchange : secret -> Cstruct.t -> (Cstruct.t, error) result
```

And here, both `secret_of_cs` and `gen_key` return the secret itself _and_ the public (here represented as "share" - the Cstruct.t to send over the wire). The secret is a scalar - but the share/public is a Point.t -- which can be compressed or uncompressed.

hannesm:
And I have two more questions:
- is the P224 compression correct (it seems it is very independent of pident)
- Where is the compression specified? Is there an IETF RFC, an ANSI specification, or something else?

hannesm:
DH API changses in https://github.com/hannesm/mirage-crypto/commit/363a18e4f24127124a5454510c49bbea6afb178d

Zett98:
> And I have two more questions:
> 
> * is the P224 compression correct (it seems it is very independent of pident)
> * Where is the compression specified? Is there an IETF RFC, an ANSI specification, or something else?
- 1  Yes     
- 2 Compression method -> https://www.secg.org/sec1-v2.pdf section 2.3, decompression method  I used  -> https://tools.ietf.org/id/draft-jivsov-ecc-compact-00.xml#def section 4.3

hannesm:
thanks again, merged as #147

PR #148, opened 2021-12-05
Debian fixes

glondu:
Here are patches, applied in Debian, to improve portability of mirage-crypto.

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -135,6 +135,14 @@ static inline uint64_t getticks(void)
 }
 #endif
 
+#if defined (__mips__)
+static inline unsigned long get_count() {
+  unsigned long count;
+  __asm__ __volatile__ ("rdhwr %[rt], $2" : [rt] "=d" (count));
+  return count;
+}
+#endif
+
 
 CAMLprim value mc_cycle_counter (value __unused(unit)) {
 #if defined (__i386__) || defined (__x86_64__)
@@ -147,6 +155,8 @@ CAMLprim value mc_cycle_counter (value __unused(unit)) {
   return Val_long (rdcycle64 ());
 #elif defined (__s390x__)
   return Val_long (getticks ());
+#elif defined(__mips__)
+  return Val_long (get_count());
 #else
 #error ("No known cycle-counting instruction.")
 #endif
--- a/src/native/ghash_ctmul.c
+++ b/src/native/ghash_ctmul.c
@@ -39,7 +39,7 @@
 #include "mirage_crypto.h"
 #include <string.h>
 
-#if defined (__i386__) || defined (__arm__)
+#if defined (__i386__) || defined (__arm__) || (defined (__mips__) && _MIPS_SIM==_ABIO32)
 
 /*
  * We cannot really autodetect whether multiplications are "slow" or
--- a/src/native/mirage_crypto.h
+++ b/src/native/mirage_crypto.h
@@ -47,9 +47,9 @@ extern struct _mc_cpu_features mc_detected_cpu_features;
 
 #endif /* __mc_ACCELERATE__ */
 
-#if defined (__x86_64__) || defined (__aarch64__) || defined (__powerpc64__) || (64 == __riscv_xlen) || defined (__s390x__)
+#if defined (__x86_64__) || defined (__aarch64__) || defined (__powerpc64__) || (64 == __riscv_xlen) || defined (__s390x__) || (defined (__mips__) && _MIPS_SIM==_ABI64)
 #define ARCH_64BIT
-#elif defined (__i386__) || defined (__arm__) || (32 == __riscv_xlen)
+#elif defined (__i386__) || defined (__arm__) || (32 == __riscv_xlen) || (defined (__mips__) && _MIPS_SIM==_ABIO32)
 #define ARCH_32BIT
 #else
 #error "unsupported platform"


glondu:
A [question](https://lists.debian.org/debian-mips/2021/12/msg00009.html) from a MIPS porter:

> what's the requirement for this cycle count ? Is it about more fine grained timestamps or really cpu cylces done ?

hannesm:
Thanks for your PR. Are you aware of any CI system with MIPS, so once we merge these changes we ensure to not break it again in the future?

> what's the requirement for this cycle count ? Is it about more fine grained timestamps or really cpu cylces done ?

The purpose is entropy harvesting - the lower 32 bit of mc_cycle_counter are mixed into the RNG (Fortuna). This is done in Lwt and MirageOS systems and called on each entry to the Lwt loop. The requirement is "a value that is hardly predictable and should be different in each call".

hannesm:
FWIW, chromium (which inspired e.g. the powerpc64 code path) - see https://chromium.googlesource.com/external/gperftools/+/master/src/base/cycleclock.h:

```
#elif defined(__mips__)
    // mips apparently only allows rdtsc for superusers, so we fall
    // back to gettimeofday.  It's possible clock_gettime would be better.
    struct timeval tv;
    gettimeofday(&tv, NULL);
    return static_cast<int64>((tv.tv_sec + tv.tv_usec * 0.000001)
                              * CyclesPerSecond());
```

Since I neither have a MIPS platform nor I am knowledgable about it, what does `   __asm__ __volatile__ ("rdhwr %[rt], $2" : [rt] "=d" (count));` actually return? Will two subsequent calls lead to a different number (could you try running `test_entropy.exe` multiple times and see if it outputs `test entropy OK`?).

hannesm:
Thanks again for your PR. I picked the first and last commit and pushed them to the main branch.

The MIPS support: unless we get a CI system, I'm not sure how to maintain platform support on MIPS (and additionally, I'm not sure about the code added there, see above question). While adding more platforms is great, it needs to have a maintenance story and a security story.

hannesm:
I added a comment to mc_cycle_counter in https://github.com/mirage/mirage-crypto/commit/bd61c160a136d88c8d450abdf1866381b4acef55 -- hopefully this clears things up for the future.

glondu:
> Since I neither have a MIPS platform nor I am knowledgable about it, what does ` __asm__ __volatile__ ("rdhwr %[rt], $2" : [rt] "=d" (count));` actually return?

According to [this](https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00086-2B-MIPS32BIS-AFP-05.04.pdf), it returns a "high-resolution cycle counter".

> Will two subsequent calls lead to a different number (could you try running `test_entropy.exe` multiple times and see if it outputs `test entropy OK`?).

Here is the output when I run this executable multiple times:
```
(sid_mipsel-dchroot)glondu@eller:~/ocaml-mirage-crypto-0.10.5$ dune exec -- tests/test_entropy.exe
no CPU RNG available 
test entropy OK
(sid_mipsel-dchroot)glondu@eller:~/ocaml-mirage-crypto-0.10.5$ dune exec -- tests/test_entropy.exe
no CPU RNG available 
test entropy OK
(sid_mipsel-dchroot)glondu@eller:~/ocaml-mirage-crypto-0.10.5$ dune exec -- tests/test_entropy.exe
no CPU RNG available 
test entropy OK
(sid_mipsel-dchroot)glondu@eller:~/ocaml-mirage-crypto-0.10.5$ dune exec -- tests/test_entropy.exe
no CPU RNG available 
test entropy OK
(sid_mipsel-dchroot)glondu@eller:~/ocaml-mirage-crypto-0.10.5$ dune exec -- tests/test_entropy.exe
no CPU RNG available 
test entropy OK
```
Cannot this test be run automatically with `dune runtest -p mirage-crypto-rng`? This command is run as part of the package building process.

glondu:
> Thanks for your PR. Are you aware of any CI system with MIPS, so once we merge these changes we ensure to not break it again in the future?

I am not aware of such a system that would be freely accessible to anyone. I guess one can be built with actual hardware, or QEMU.

Debian has some mips{,64}el machines dedicated to package building, and a porterbox (only Debian Developers have access to it) where I did my work. Our policy is to run as much tests as possible during package building, so if something is broken, it can be spot this way. So if you are afraid of breaking something, put a test for it in the testsuite.

glondu:
> The MIPS support: unless we get a CI system, I'm not sure how to maintain platform support on MIPS (and additionally, I'm not sure about the code added there, see above question). While adding more platforms is great, it needs to have a maintenance story and a security story.

At the moment, mirage-crypto has a terrible portability story:

  https://buildd.debian.org/status/package.php?p=ocaml-mirage-crypto&suite=unstable

For this reason, I am reluctant to start using it in my own projects.

I don't understand the will to restrict portability in such a drastic way. For example, concerning rng, why insisting on having a CPU-specific cycle counter? Why isn't a call to getrandom enough (at least on Linux)?

hannesm:
Dear @glondu,

thanks for your comments and runs.

> Cannot this test be run automatically with dune runtest -p mirage-crypto-rng? This command is run as part of the package building process

This is the case, but since I don't use debian and do not know about what packagers do, I was not aware that you run the testsuite.

>  >  Since I neither have a MIPS platform nor I am knowledgable about it, what does __asm__ __volatile__ ("rdhwr %[rt], $2" : [rt] "=d" (count)); actually return?
>
> According to this, it returns a "high-resolution cycle counter".

Thanks for the reference.

> For example, concerning rng, why insisting on having a CPU-specific cycle counter? Why isn't a call to getrandom enough (at least on Linux)?

One of the motivations of mirage-crypto (earlier nocrypto) is indeed MirageOS - a library operating system running on hypervisors. In this setting, `getrandom` is not supported, but we want to collect some entropy, and hooking a high-resolution counter in the event loop is a promising source (of low-quality entropy).

> At the moment, mirage-crypto has a terrible portability story:
>
> https://buildd.debian.org/status/package.php?p=ocaml-mirage-crypto&suite=unstable
>
> For this reason, I am reluctant to start using it in my own projects.

I understand your reluctance, and the question is person power and usage of such a library. I only have access to some architectures, and interest in some platforms. I'm fine with increased portability, but as outlined above I'm keen to not break support for platforms by innocently looking commits (and this is why a CI for each platform is very nice to have).

From a different perspective, mirage-crypto is supported on both big-endian and little-endian systems, 32 and 64 bit -- on widely deployed platforms (x86, arm, amr64), even on new ones (risc-v, s390x). The major work for supporting new platforms (s390 cb2c26463195414f312fe96cf79501ac46e90e7a) is figuring out a cycle_counter, and adding one or two ifdefs (for figuring out whether the architecture is 32 bit or 64 bit).

Tl;DR: I'm fine to merge this with the additional information you provided, plus your test output. Thanks a lot.

hannesm:
Thanks, merged as f7f6cec96a47ee330285f48dd74324f68960a441

PR #150, opened 2022-02-01
On Windows, run `PKG_CONFIG_PATH` through `cygpath`

MisterDA:
When using OCaml in Windows, `PKG_CONFIG_PATH` is usually in Windows
style. GNU Make with the `shell` function will call Cygwin's Unix shell
which doesn't recognize Windows paths. If we're on a Windows system,
and if we find `cygpath`, convert the path list to Unix style.

This is the same patch that I've pushed to mirage-tcpip and elsewhere.


MisterDA:
I don't know why the CI on Windows OCaml 4.09 is failing, but it's not related to this patch.
ping @hannesm, @dinosaure

hannesm:
yes, this patch looks good. would it be fine with you to defer that until we've mirage4 out and can remove these Makefiles completely?

MisterDA:
Yep, I'll just pin the package.

PR #151, opened 2022-02-02
adapt to mirage-unix 5.0.0

hannesm:
//cc @dinosaure

### DIFF
--- a/mirage-crypto-rng-mirage.opam
+++ b/mirage-crypto-rng-mirage.opam
@@ -23,7 +23,7 @@ depends: [
   "mirage-runtime" {>= "3.8.0"}
   "mirage-time" {>= "2.0.0"}
   "mirage-clock" {>= "3.0.0"}
-  "mirage-unix" {with-test & >= "3.0.0"}
+  "mirage-unix" {with-test & >= "5.0.0"}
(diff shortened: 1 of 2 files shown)


hannesm:
this results in ocaml-ci being stuck that mirage-runtime > 4.0 is now in the (test) dependency chain which is only available in opam-version >= 2.1. //cc @talex5 any chance we can get ocaml-ci with opam 2.1?

talex5:
@samoht merged a PR that was supposed to fix that: https://github.com/ocurrent/ocaml-ci/pull/413

dinosaure:
For me it's perfect. We don't need a release of the distribution when the `mirage-unix` is bring only by testing.

hannesm:
@talex5 @samoht thanks, so but according to the CI log this has not been fixed... maybe it makes a difference between an opam file with availability vs a transitive dependency that has this availability?

talex5:
I think it's just a bug parsing the string: https://github.com/ocurrent/ocaml-ci/issues/412#issuecomment-1029128460

hannesm:
thanks to everyone involved, the ocaml-ci seems to be successful now (also on non-amd64 architectures).

PR #152, opened 2022-03-29
ocaml-solo5 defines __ocaml_solo5__

hannesm:
//cc @dinosaure @samoht -- is this the right thing to do!?


dinosaure:
Yes, this is what I did for `digestif` :+1:.

PR #155, opened 2022-04-22
mirage-crypto-rng-eio: Eio backend of mirage-crypto-rng

bikallem:
This PR introduces a new opam package `mirage-crypto-rng-eio`. The purpose of this package is to implement an [eio](https://github.com/ocaml-multicore/eio) dependent random number generator. 

It closely follows `mirage-crypto-rng.unix` library.


hannesm:
Thanks for your work. I've no good understanding of "eio", but doesn't that provide tasks / threads or effects and timers? Wouldn't it be more sensible to integrate a continuous (periodic) reseeding approach, as done in lwt and mirage and async implementations?

bikallem:
Indeed `eio` does have task,fibers and timers. Let me have a look at lwt implementation.

hannesm:
And while at it, please describe in more detail what
```OCaml
  let rand_source = Eio.Stdenv.secure_random env in
  let got = Eio.Flow.read rand_source buf in
```

Does -- i.e. where does it take its entropy from? Does it work for arbitrary big buffers `buf`? What is the failure semantics thereof? For the unix implementation, `getrandom` is used (see `mc_getrandom_stubs.c`) and the respective failure modes (i.e. handline of `EINTR`, calling `getrandom` again if a smaller size was returned).

Also, the CI seems to fail "cannot find package eio".

bikallem:
For `eio_linux`, the entropy is same as `mirage-crypto-rng.unix` - `get_random` https://github.com/ocaml-multicore/eio/blob/main/lib_eio_linux/eio_stubs.c#L33. 

For other OS, the entropy source is as defined in libuv lib - http://docs.libuv.org/en/v1.x/misc.html#c.uv_random

The buffers are abitrary.

`EINTR` is not handled in the eio lib itself - at least for eio_linux. I need to check if libuv handles `EINTR` or not. Let me come back to you on this later. 

The eio package is here - https://opam.ocaml.org/packages/eio/. Perhaps the opam repo needs to be updated?

hannesm:
Hmm, looks like there are multiple issues:

> For eio_linux, the entropy is same as mirage-crypto-rng.unix - getrandom https://github.com/ocaml-multicore/eio/blob/main/lib_eio_linux/eio_stubs.c#L33.

There are differences between these two C implementations, esp. if getrandom returned a short read.

> For other OS, the entropy source is as defined in libuv lib - http://docs.libuv.org/en/v1.x/misc.html#c.uv_random

*shrug* maybe

> The buffers can be abitrary/user-defined size.

That sounds good.

> EINTR is not handled in the eio lib itself - at least for eio_linux. I need to check if libuv handles EINTR or not. Let me come back to you on this later.

Ok

> The eio package is here - https://opam.ocaml.org/packages/eio/. Perhaps the opam repo needs to be updated?

Sorry, I don't understand "perhaps the opam repo needs to be updated". The "ocaml-ci" run here now only executes on the "debian-11-4.12+domains" (and reutrns [(failed: Library "eio" not found.)](https://ci.ocamllabs.io/github/mirage/mirage-crypto/commit/88eb7c82ff443583f32d8ac906c87cdf91fd76a6/variant/debian-11-4.12+domains)). This is a regression to earlier runs, where lots of compilers and platforms were used. Maybe an issue with ocaml-ci (with multiple opam packages in a ssingle repository), and definitively a blocker to merge this PR.

I'm sorry I won't have time to look into further details before mid May (or even early June),

rng/eio/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,15 @@
...
+
+(** [initialize ~sleep env sw] will bring the RNG into a working state. The argument
+    [sleep] is measured in ns (default 1s), and used to sleep between collection
+    of entropy from the CPU RNG, every [10 * sleep] getrandom is used to collect
+    entropy.
+*)
+val initialize : ?sleep:int64 -> Eio.Stdenv.t -> Eio.Switch.t -> unit 
+
+(** [getrandom env size] returns a buffer of [size] filled with random bytes. *)
+val getrandom: Eio.Stdenv.t -> int -> Cstruct.t
```
talex5, line 15: This only really needs the `secure_random` flow, not the whole stdenv; the inferred type is:
```suggestion
val getrandom: < secure_random : #Eio.Flow.source; .. > -> int -> Cstruct.t
```
It would be simpler to take the flow directly though:
```suggestion
val getrandom: secure_random:#Eio.Flow.source -> int -> Cstruct.t
```
However, since this function doesn't interact with mirage-crypto, perhaps it just shouldn't be exposed at all? Users can read from the flow themselves quite easily.

rng/eio/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let getrandom env i = 
+  let buf = Cstruct.create i in
+  let rand_source = Eio.Stdenv.secure_random env in
+  let got = Eio.Flow.read rand_source buf in
+  if got < i then 
+    failwith (Printf.sprintf "get_random_init: expected %d \
+                              but got %d bytes of random data" i got)
+  else 
+    buf
```
talex5, line 14: Is it important that it all comes in a single read? If not:

```suggestion
  Eio.Flow.read_exact rand_source buf;
  buf
```

talex5 (commented):
Thanks - I've filed an issue about EINTR with getrandom https://github.com/ocaml-multicore/eio/issues/211.

> The "ocaml-ci" run here now only executes on the "debian-11-4.12+domains"

Yes, this is a problem with the CI. I think we should fix this before trying to upstream any eio support (https://github.com/ocurrent/ocaml-ci/issues/297).

@bikallem: it might be convenient to have a `Mirage_crypto_rng_eio.run` function, to avoid the need for a switch here. e.g.

```ocaml
Eio_main.run @@ fun env ->
Mirage_crypto_rng_eio.run env @@ fun () ->
...
```

bikallem:
Thanks for the review.
> it might be convenient to have a Mirage_crypto_rng_eio.run function, to avoid the need for a switch here. e.g.

Do you mean I do `Eio.Switch.run` inside `Mirage_crypto_rng_eio.run`?

bikallem:
@talex5 with regards to handling EINTR for getrandom, this has to be done separately in both `eio_linux` and `eio_luv` right?

talex5:
> Do you mean I do Eio.Switch.run inside Mirage_crypto_rng_eio.run?

I was thinking of something like this:
```ocaml
let run ?(sleep = Duration.of_sec 1) env fn =
  if !running then (
    Log.debug
      (fun m -> m "Mirage_crypto_rng_eio.initialize has already been called, \
                   ignoring this call.");
    fn ()
  ) else (
    running := true;
    Fun.protect ~finally:(fun () -> running := false) @@ fun () ->
    (try
      let _ = default_generator () in
      Log.warn (fun m -> m "Mirage_crypto_rng.default_generator has already \
                            been set, check that this call is intentional");
     with 
      No_default_generator -> ());
    let seed = 
      let init = 
        Entropy.[ bootstrap ; whirlwind_bootstrap ; bootstrap ; getrandom_init env ] in
      List.mapi (fun i f -> f i) init |> Cstruct.concat
    in 
    let rng = create ~seed ~time:Mtime_clock.elapsed_ns (module Fortuna) in
    set_default_generator rng;
    let source = Entropy.register_source "getrandom" in (* xxx: no way to unregister? *)
    let feed_entropy () = periodically_feed_entropy env (Int64.mul sleep 10L) source in
    Eio.Fiber.any (rdrand_tasks env sleep @ [feed_entropy; fn])
  )
```
I used `any` instead of a switch so that the rng will stop when the main function exits.

bikallem:
> > Do you mean I do Eio.Switch.run inside Mirage_crypto_rng_eio.run?
> 
> I was thinking of something like this:
> 
> ```ocaml
> let run ?(sleep = Duration.of_sec 1) env fn =
>   if !running then (
>     Log.debug
>       (fun m -> m "Mirage_crypto_rng_eio.initialize has already been called, \
>                    ignoring this call.");
>     fn ()
>   ) else (
>     running := true;
>     Fun.protect ~finally:(fun () -> running := false) @@ fun () ->
>     (try
>       let _ = default_generator () in
>       Log.warn (fun m -> m "Mirage_crypto_rng.default_generator has already \
>                             been set, check that this call is intentional");
>      with 
>       No_default_generator -> ());
>     let seed = 
>       let init = 
>         Entropy.[ bootstrap ; whirlwind_bootstrap ; bootstrap ; getrandom_init env ] in
>       List.mapi (fun i f -> f i) init |> Cstruct.concat
>     in 
>     let rng = create ~seed ~time:Mtime_clock.elapsed_ns (module Fortuna) in
>     set_default_generator rng;
>     let source = Entropy.register_source "getrandom" in (* xxx: no way to unregister? *)
>     let feed_entropy () = periodically_feed_entropy env (Int64.mul sleep 10L) source in
>     Eio.Fiber.any (rdrand_tasks env sleep @ [feed_entropy; fn])
>   )
> ```
> 
> I used `any` instead of a switch so that the rng will stop when the main function exits.

I was thinking more in terms of API ergonomics not having to do the following:
```
Eio_main.run @@ fun env ->
Eio.Switch.run @@ fun _sw ->
...
```
Perhaps, a version of `Eio_main.run` as `Eio_main.run : ?sw:Switch.t -> (Eio.Stdenv.t -> Eio.Switch.t -> unit) -> unit` ? could be useful in such scenarios?

talex5:
You need to know when the user's main function returns so that the rng fiber can be stopped too. `Fiber.any` will do this, but a switch won't. There's shouldn't be any need for the user to create a switch here, they just do:

```ocaml
let () =
  Eio_main.run @@ fun env ->
  Mirage_crypto_rng_eio.run env @@ fun () ->
  main ()     (* Can use mirage_crypto in here *)
```

bikallem:
> You need to know when the user's main function returns so that the rng fiber can be stopped too. `Fiber.any` will do this, but a switch won't. There's shouldn't be any need for the user to create a switch here, they just do:
> 
> ```ocaml
> let () =
>   Eio_main.run @@ fun env ->
>   Mirage_crypto_rng_eio.run env @@ fun () ->
>   main ()     (* Can use mirage_crypto in here *)
> ```

Ah, yes. I get it now. The API looks nice

bikallem:
@hannesm 
>EINTR is not handled in the eio lib itself - at least for eio_linux. I need to check if libuv handles EINTR or not. Let me come back to you on this later.

>Ok - please have a look and report back (this is a blocker for merging).

EINTR is now handled correctly by eio. https://github.com/ocaml-multicore/eio/pull/212

hannesm:
Thanks for your work. I did not follow the discusion, but as you may see the CI is not happy. And as remarked earlier, please make sure that the CI continues to run on x86_32 etc. (see https://github.com/mirage/mirage-crypto/runs/5740484133 for what the CI should run on).

> EINTR is now handled correctly by eio. https://github.com/ocaml-multicore/eio/pull/212

You may need to cut a release of `eio` and add some lower bounds in the dependencies here.

In the current state this is not mergeable. I will try to look into this PR again (likely) early June.

rng/eio/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,22 @@
...
+
+    [[
+      let () =
+        Eio_main.run @@ fun env ->
+        Mirage_crypto_rng_eio.run env @@ fun () -> 
+        let randon_num = Mirage_crypto_rng.generate 32 in 
+        Printf.printf "Random number: %S%!" (Cstruct.to_string randon_num)
+    ]]
+*)
+val run : ?sleep:int64 -> Eio.Stdenv.t -> (unit -> unit) -> unit
```
talex5, line 22: It would be nice to pass through the user's return value:
```suggestion
val run : ?sleep:int64 -> Eio.Stdenv.t -> (unit -> 'a) -> 'a
```
This will also act as a check that the other fibers can't ever finish normally (they can only be cancelled), since they won't be able to return an `'a`. For example, it will uncover the bug in `rdrand_task` where it cancels the application if rdrand isn't available, and the bug in the `if !running` branch. The code examples I posted earlier in the comments have fixes for these problems.

I think it would also be nice to be explicit about which parts of `Stdenv.t` are used. This also makes it easier to use if you have a richer environment (e.g. from `Eio_linux.run`, which adds Linux-specific features), or to use it from unit-tests.

```suggestion
type env = <
  clock : Eio.Time.clock;  (* Used to mix in extra randomness periodically *)
  secure_random : Eio.Flow.source;  (* OS source of extra randomness *)
>

val run : ?sleep:int64 -> <env; ..> -> (unit -> unit) -> unit 
```

talex5 (commented):
> This is a regression to earlier runs, where lots of compilers and platforms were used. Maybe an issue with ocaml-ci (with multiple opam packages in a ssingle repository), and definitively a blocker to merge this PR.

I deployed a fix for this (https://github.com/ocurrent/ocaml-ci/pull/468). However, it's not accepted into ocaml-ci yet, so might get reverted.

hannesm:
@bikallem in theory yes. In practise there are several issues
- CI is failing
- I can't see any tests that exercise the code

In addition, since @talex5 did a review and various suggestions, I trust him that the code is fine. So, as stated earlier, please look in depth in the CI failures (of all the CI systems, from GitHub actions over cirrus to OCaml CI). For comparison, the CI on the main branch succeeds (and in OCaml-CI there are various architectures being tested, ...).

Please do not put my name in the authors or maintainers list of the opam package of "mirage-crypto-rng-eio".

For testing, maybe test_entropy_collection and test_entropy_collection_async can provide you with some inspiration what would be sensible to test.

I'm on vacation at the moment, and may take a look mid-July.

bikallem:
@hannesm There is a PR to address some of the CI errors (https://github.com/mirage/mirage-crypto/pull/157). Could we merge this PR. Once that is merged, I will look into adding github actions and cirrus support for the `eio` packages.

dinosaure:
> @hannesm There is a PR to address some of the CI errors (https://github.com/mirage/mirage-crypto/pull/157). Could we merge this PR. Once that is merged, I will look into adding github actions and cirrus support for the eio packages.

I think you have reversed the order of the PRs. `mirage-crypto` is a sensitive repository in the merge policy is very strict. It would be wise to fix the CI problem with respect to `eio` (especially about `ocaml-ci`) first. Only then, we will be able to merge this PR.

bikallem:
> > @hannesm There is a PR to address some of the CI errors (#157). Could we merge this PR. Once that is merged, I will look into adding github actions and cirrus support for the eio packages.
> 
> I think you have reversed the order of the PRs. `mirage-crypto` is a sensitive repository in the merge policy is very strict. It would be wise to fix the CI problem with respect to `eio` (especially about `ocaml-ci`) first. Only then, we will be able to merge this PR.

Thanks, noted. I have closed that PR now. I will do the ci fixes/improvements on this PR.

bikallem:
@hannesm @dinosaure The cirrus CI seems to be generally working. However, it fails on `./.test-mirage.sh` due to opam issue I think,
```
# Detecting depexts using vars: arch=x86_64, os=freebsd, os-distribution=freebsd, os-family=bsd
[ERROR] No solution for mirage-unikernel-crypto_test-unix:   * Missing dependency:
            - mirage-crypto-rng-mirage < 0.11.0
            not available because the package is pinned to version dev
```
Looks like this is mostly an opam issue ?Any pointers?

bikallem:
/cc @dinosaure @hannesm 
All CIs are now green. ocaml-ci, cirrus and gha(https://github.com/bikallem/mirage-crypto/actions)

bikallem:
Added entropy test as suggested. 

The PR is ready for approval/final review now.

.cirrus.yml
```diff
@@ -1,14 +1,42 @@
 freebsd_instance:
   image_family: freebsd-12-2
 
-freebsd_task:
+freebsd_default_task:
```
hannesm, line 4: what is the purpose of the rename?

.cirrus.yml
```diff
@@ -1,14 +1,42 @@
...
+    - opam env
+  
+  pin_packages_script: 
+    - opam install -y solo5-bindings-hvt zarith-freestanding opam-depext
+    - opam pin add mirage-crypto . -y --with-version=0.10.6  --with-test
+    - opam pin add mirage-crypto-rng . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-rng-mirage . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-pk . -y  --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-ec . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-rng-async . -y --with-version=0.10.6 --with-test
```
hannesm, line 23: I don't understand the changes above. Could you minimize the difference in this file, please?

hannesm:
Thanks for your PR, would you mind to minimize the changes to existing CI scripts?

> Github Actions all are passing now. I have added two additional compilers 4.13.1 and 4.14.0.
> Cirrus CI seems to be working except for the issue below. I have upgraded the compilers to 4.13.1 and 4.14.0.

Would you mind to separate the changes you intend to do on the existing CI scripts (such as renaming files, "upgrading compilers" etc.) to a separate PR, please?

.cirrus.yml
bikallem: Mostly to distinguish tasks between default - all other compilers vs `5.0.0` (eio capable) compiler.

.cirrus.yml
```diff
@@ -1,14 +1,42 @@
...
   image_family: freebsd-12-2
 
-freebsd_task:
+freebsd_default_task:
   env:
     matrix:
-      - OCAML_VERSION: 4.11.1
-      - OCAML_VERSION: 4.12.0
+      - OCAML_VERSION: 4.13.1
+      - OCAML_VERSION: 4.14.0
```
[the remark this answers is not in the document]
bikallem: I assumed the repo. kept up to date with the latest last 2 OCaml releases. If it is not desired I will revert it.

.github/workflows/unix.yml
```diff
@@ -0,0 +1,70 @@
+name: Unix
```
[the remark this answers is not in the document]
bikallem: ok. will revert.

.github/workflows/windows.yml
```diff
@@ -24,13 +23,14 @@ jobs:
...
             !mirage-crypto-rng-async.opam
+            !mirage-crypto-rng-eio.opam
           ocaml-compiler: ${{ matrix.ocaml-version }}
 
       - name: Install dependencies
         run: opam install --deps-only -t mirage-crypto mirage-crypto-rng mirage-crypto-rng-mirage mirage-crypto-pk mirage-crypto-ec
 
       - name: Build
-        run: opam exec -- dune build -p mirage-crypto,mirage-crypto-rng,mirage-crypto-rng-mirage,mirage-crypto-pk,mirage-crypto-ec
+        run: opam exec -- dune build -p mirage-crypto,mirage-crypto-rng,mirage-crypto-rng-mirage,mirage-crypto-pk,mirage-crypto-ec --profile release
```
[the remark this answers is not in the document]
bikallem: GHA `setup-ocaml` tries to build all packages specified in `opam-local-packages` with all of the OCaml compilers specified in `ocaml-version`. This will result in error since 4.* series of OCaml compilers can't build `mirage-crypto-rng-eio` pakage. Therefore we need to exclude the package from being built by 4.* series OCaml compilers. 

--profile release is strictly not necessary but sometimes `dune runtest` is bit faster and saves some CI time. 

I have added the latest 2 OCaml release compilers just to be sure we are compatible with those versions. However, this can be removed if there is no desire for it. It does add CI time.

.cirrus.yml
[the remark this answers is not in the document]
bikallem: Reverted the change.

.github/workflows/windows.yml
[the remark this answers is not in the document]
bikallem: reverted.

bikallem:
@hannesm I believe I have addressed all review comments regarding CI changes.

bikallem:
ping @hannesm @dinosaure . I think the failed ci for OCaml 4.11.2 on windows in non related to the PR. It seems to be spurious. The GHA CI seems to pass in my branch (https://github.com/bikallem/mirage-crypto/actions)

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
...
+    - opam update
+    - opam switch create 5.0.0~alpha0 --repositories=default,alpha=git+https://github.com/kit-ty-kate/opam-alpha-repository.git
+    - opam env
+
+  pin_packages_script: 
+    - opam pin add mirage-crypto . -y --with-version=dev
+    - opam pin add mirage-crypto-rng . -y --with-version=dev
+    - opam pin add mirage-crypto-rng-eio . -y --with-version=dev --with-test
+
+  test_script: opam exec -- dune runtest eio --profile release
```
hannesm, line 42: Could you try to minimize the difference please, and/or explain why `--profile release` is used here?

.github/workflows/test.yml
```diff
@@ -21,13 +21,51 @@ jobs:
...
+          ocaml-compiler: ${{ matrix.ocaml-version }}
+          opam-repositories: |
+            default: https://github.com/ocaml/opam-repository.git
+            alpha: https://github.com/kit-ty-kate/opam-alpha-repository.git
+
+      - name: Install dependencies
+        run: opam install --deps-only -t mirage-crypto mirage-crypto-rng mirage-crypto-rng-eio
+
+      - name: Build
+        run: opam exec -- dune build -p mirage-crypto,mirage-crypto-rng,mirage-crypto-rng-eio --profile release
```
hannesm, line 68: Why the `--profile release`?

.github/workflows/test.yml
```diff
@@ -21,13 +21,51 @@ jobs:
...
+
+      - name: Test
+        run: opam exec -- dune runtest eio --profile release
```
hannesm, line 71: Why the `--profile release`? Why `dune runtest eio`? Would a `dune runtest` not work?

hannesm:
Dear @bikallem. many thanks for your contribution. Sorry for being slow and picky on this PR, but adding more code (more features) to mirage-crypto means having to maintain more code. And especially divergence in the CI scripts is something I need to be aware of for future changes.

I reviewed your changes to the CI scripts again, and I have several questions (see the comments above).

Please also remember that I maintain this code in my spare time. Commenting multiple times a week "ping" actually lowers my motivation to look into the PR.

.github/workflows/test.yml
bikallem: Removed `--profile release`. `dune runtest` results in build error.

.github/workflows/test.yml
bikallem: removed now.

.github/workflows/test.yml
```diff
@@ -21,13 +21,51 @@ jobs:
...
-        run: opam install --deps-only -t .
+        run: opam install --deps-only -t mirage-crypto mirage-crypto-rng mirage-crypto-rng-mirage mirage-crypto-pk mirage-crypto-ec mirage-crypto-rng-async
 
       - name: Build
-        run: opam exec -- dune build
+        run: opam exec -- dune build -p mirage-crypto,mirage-crypto-rng,mirage-crypto-rng-mirage,mirage-crypto-pk,mirage-crypto-ec,mirage-crypto-rng-async --profile release
 
       - name: Test
-        run: opam exec -- dune runtest
+        run: opam exec -- dune runtest tests --profile release
```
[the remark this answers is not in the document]
bikallem: `dune runtest` results in error since it can't build `eio` package. So we need to be specific here. Removed `--profile release`.

.cirrus.yml
bikallem: It makes the tests a bit faster since the resulting executable is optimized.

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
...
+  pin_packages_script: 
+    - opam install -y solo5-bindings-hvt zarith-freestanding opam-depext
+    - opam pin add mirage-crypto . -y --with-version=0.10.6  --with-test
+    - opam pin add mirage-crypto-rng . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-rng-mirage . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-pk . -y  --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-ec . -y --with-version=0.10.6 --with-test
+    - opam pin add mirage-crypto-rng-async . -y --with-version=0.10.6 --with-test
+
+  test_script: opam exec -- dune runtest tests --profile release
```
hannesm, line 25: Again, can we please drop `--profile release` to see errors early in CI?

When you comment with "It makes the tests a bit faster since the resulting executable is optimized.", would you mind to share actual numbers if you think it is worth including that change.

.cirrus.yml
hannesm, line 42: See above.

.cirrus.yml
bikallem: removed now.

.cirrus.yml
bikallem: removed.

mirage-crypto-rng-eio.opam
```diff
@@ -0,0 +1,29 @@
...
+synopsis:     "Feed the entropy source in an eio-friendly way"
+
+build: [ ["dune" "subst"] {dev}
+         ["dune" "build" "-p" name "-j" jobs ]
+         ["dune" "runtest" "-p" name "-j" jobs] {with-test} ]
+
+depends: [
+  "base-domains"
+  "dune" {>= "3.0"}
+  "eio"
```
hannesm, line 18: what about the comment from May 9th:

> > EINTR is now handled correctly by eio. https://github.com/ocaml-multicore/eio/pull/212
>
> You may need to cut a release of eio and add some lower bounds in the dependencies here.

Would you mind to add a reasonable lower bound here, to avoid random people running into (rare) errors that can be avoided by using a recent `eio`.

bikallem: Done.

hannesm:
Ok, some final questions:
- @talex5 since you know a lot of eio, does this PR look fine to you?
- Is there - similar to lwt and async - a way to execute some code whenever a fiber (domain? task?) is scheduled? (in Lwt, we use Lwt_main.Enter_iter_hooks.add_first, in async Scheduler.Expert.run_every_cycle_start). This would be great to add some further bits of entropy to the pools (now it is only Eio.Stdenv.secure_random and rdrand (if supported by CPU))

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,78 @@
...
+  = 
+    if !running then begin 
+      Log.debug
+        (fun m -> m "Mirage_crypto_rng_eio.initialize has already been called, \
+                     ignoring this call.");
+      fn () 
+    end
+    else begin
+      running := true;
+      Fun.protect ~finally:(fun ()-> running := false) @@ fun () ->
```
hannesm, line 61: could you rewrite without using `@@` (which imho is hard to keep in the brain, much easier to use `begin` .. `end` or `(` `)`.

hannesm: also, please add a whitespace before the arrow (`fun () -> running := false`).

bikallem: Done.

eio/tests/test_eio_rng.ml
```diff
@@ -0,0 +1,11 @@
+open Mirage_crypto_rng
+
+let () = 
+  Eio_main.run @@ fun env -> 
+   Mirage_crypto_rng_eio.run (module Fortuna) env @@ fun () ->
```
talex5, line 5:
```suggestion
  Mirage_crypto_rng_eio.run (module Fortuna) env @@ fun () ->
```

bikallem: done.

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+    [sleep] is measured in ns (default 1s), and used to sleep between collection
+    of entropy from the CPU RNG, every [10 * sleep] getrandom is used to collect
+    entropy.
+
+    [fn] is the main function that will have access to a running RNG.
+
+    [[
+      let () =
+        Eio_main.run @@ fun env ->
+        Mirage_crypto_rng_eio.run env @@ fun () ->
```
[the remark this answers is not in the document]
hannesm: Thanks @talex5 for your comment, I opened #160 to track this inconsistency.

I also revisited how entropy feeding and the default_generator should behave (and pushed 3871e3b for streamlining and explanation).

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+        let random_num = Mirage_crypto_rng.generate 32 in 
+        Printf.printf "Random number: %S%!" (Cstruct.to_string random_num)
+    ]]
+*)
+val run 
+  :  ?g:'a
+  -> ?sleep:int64 
+  -> 'a Mirage_crypto_rng.generator
+  -> <env; ..> 
+  -> (unit -> 'b) -> 'b
```
hannesm, line 32: After thinking again about 3871e3b, I wonder what the semantics of `run` is (and why it is as it is). According to the documentation, it accepts as last argument a function `fn` "that will have access to a running RNG".

So is it expected that `run` is executed multiple times (on different domains)?

While it sets and resets the module-local mutable state `running`, a second execution will print a warning message. Also, if I understand the implementation correctly (please help me since I'm a bit lost in respect to cancellation and eio), after `run` has been executed (and finished), any `Mirage_crypto_rng.generate` will produce random numbers. But once the `fn` in `run` finishes, the periodic entropy feeding is stopped (and thus the rng keeps to produce random numbers, but that opens attacks if some bits of the rng state can be predicted -- this is the reason why periodic entropy feeding is necessary).

The other implementations (lwt, mirage, async) expect `initialize` (and why is it called `run` here?) to be called exactly once, and this `initialize` registers several tasks, and finishes immediately -- but the entropy feeding is never stopped.

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+    This module initializes a Fortuna RNG with [getrandom()], and CPU RNG.
+    [Eio.Stdenv.secure_random] is used as the [getrandom()] implementation.
+*)
+
+type env = <
+  clock: Eio.Time.clock;
+  secure_random: Eio.Flow.source;
+  >
+
+(** [run ~sleep env fn] will bring the RNG into a working state. The argument
```
hannesm, line 12: This docstring is not congruent with the interface below: `~g` and `generator` are missing.

eio/src/mirage_crypto_rng_eio.mli
hannesm: Re-reading the discussion, I noticed @talex5 is in favour of the `run` semantics "so that periodic task quits once the main task finishes". I wonder whether that should the imply that the defaiult_rng is reset to None (at the moment there's no API for this)? It is also a bit tricky, since certainly `fn` may contain the set to a different default generator, which `run` would once the execution finishes, be reset to `None`.

I'm a bit puzzled: what is the advantage of cancelling the periodic tasks? Is there an example application which only needs for a startup phase the rng? E.g. the lwt implementation registers (via Lwt.async) some tasks, so execution will be done whenever Lwt_main.run is executed -- does eio have similar functionality?

hannesm:
Another reviewing comment: There is a `setup_rng` in `rng/rng.ml` which is used for printing the Unseeded_generator and No_default_generator exceptions. This goes into detail of the different backends and should be extended with eio.

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+
+    [[
+      let () =
+        Eio_main.run @@ fun env ->
+        Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () -> 
+        let random_num = Mirage_crypto_rng.generate 32 in 
+        Printf.printf "Random number: %S%!" (Cstruct.to_string random_num)
+    ]]
+*)
+val run
```
hannesm, line 27: please avoid the trailing whitespace (after `run`)

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+      let () =
+        Eio_main.run @@ fun env ->
+        Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () -> 
+        let random_num = Mirage_crypto_rng.generate 32 in 
+        Printf.printf "Random number: %S%!" (Cstruct.to_string random_num)
+    ]]
+*)
+val run 
+  :  ?g:'a
+  -> ?sleep:int64
```
hannesm, line 29: see above, please no trailing whitespace

eio/src/mirage_crypto_rng_eio.mli
```diff
@@ -0,0 +1,32 @@
...
+        Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () -> 
+        let random_num = Mirage_crypto_rng.generate 32 in 
+        Printf.printf "Random number: %S%!" (Cstruct.to_string random_num)
+    ]]
+*)
+val run 
+  :  ?g:'a
+  -> ?sleep:int64 
+  -> 'a Mirage_crypto_rng.generator
+  -> <env; ..>
```
hannesm, line 31: please no trailing whitespace

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
     matrix:
       - OCAML_VERSION: 4.11.1
       - OCAML_VERSION: 4.12.0
+
+  pkg_install_script: pkg install -y ocaml-opam gmp gmake pkgconf bash
+
+  ocaml_script:
```
hannesm, line 12: please remove the trailing whitespace

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
...
       - OCAML_VERSION: 4.11.1
       - OCAML_VERSION: 4.12.0
+
+  pkg_install_script: pkg install -y ocaml-opam gmp gmake pkgconf bash
+
+  ocaml_script: 
+    - opam init -a --comp=$OCAML_VERSION
+    - opam env
+  
+  pin_packages_script:
```
hannesm, line 16: please remove the trailing whitespace

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
...
+
+freebsd_eio_task: 
   pkg_install_script: pkg install -y ocaml-opam gmp gmake pkgconf bash
-  ocaml_script: opam init -a --comp=$OCAML_VERSION
-  dependencies_script: eval `opam env` && opam install -y --deps-only .
-  build_script: eval `opam env` && dune build @install
-  test_script: eval `opam env` && opam install -y -t --deps-only . && dune build @runtest
-  test_mirage_script: eval `opam env` && opam install -y solo5-bindings-hvt zarith-freestanding opam-depext && opam pin add -y . && ./.test-mirage.sh
+
+  ocaml_script:
```
hannesm, line 31: please avoid the trailing whitespace

.cirrus.yml
```diff
@@ -6,9 +6,37 @@ freebsd_task:
...
-  test_script: eval `opam env` && opam install -y -t --deps-only . && dune build @runtest
-  test_mirage_script: eval `opam env` && opam install -y solo5-bindings-hvt zarith-freestanding opam-depext && opam pin add -y . && ./.test-mirage.sh
+
+  ocaml_script: 
+    - opam init -a --bare
+    - opam update
+    - opam switch create 5.0.0~alpha0 --repositories=default,alpha=git+https://github.com/kit-ty-kate/opam-alpha-repository.git
+    - opam env
+
+  pin_packages_script:
```
hannesm, line 37: please avoid the trailing whitespace

eio/src/dune
```diff
@@ -0,0 +1,10 @@
+(library
+ (name mirage_crypto_rng_eio)
+ (public_name mirage-crypto-rng-eio)
+ (libraries 
+   eio 
+   cstruct 
+   logs 
+   mirage-crypto-rng 
+   duration
```
hannesm, line 9: in lines 4-9, please remove the trailing whitespaces.

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
+open Mirage_crypto_rng
```
hannesm, line 1: trailing whitespace

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
...
+
+type env = <
+  clock: Eio.Time.clock;
+  secure_random: Eio.Flow.source;
+>
+
+let src = Logs.Src.create "mirage-crypto-rng-eio" ~doc:"Mirage crypto RNG Eio"
+module Log = (val Logs.src_log src: Logs.LOG)
+
+let getrandom env i =
```
hannesm, line 11: trailing whitespace

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
...
+
+let src = Logs.Src.create "mirage-crypto-rng-eio" ~doc:"Mirage crypto RNG Eio"
+module Log = (val Logs.src_log src: Logs.LOG)
+
+let getrandom env i = 
+  let buf = Cstruct.create i in
+  Eio.Flow.read_exact env#secure_random buf;
+  buf
+
+let getrandom_init env i =
```
hannesm, line 16: trailing whitespace

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
...
+let getrandom_init env i = 
+  let data = getrandom env 128 in
+  Entropy.header i data
+
+let rec periodic env f delta =
+  f ();
+  Eio.Time.sleep env#clock (Duration.to_f delta);
+  periodic env f delta 
+
+let periodically_feed_entropy env delta source =
```
hannesm, line 25: trailing whitespace

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
...
+  Eio.Time.sleep env#clock (Duration.to_f delta);
+  periodic env f delta 
+
+let periodically_feed_entropy env delta source = 
+  let task () = 
+    let per_pool = 8 in
+    let size = per_pool * pools None in
+    let random = getrandom env size in
+    let idx = ref 0 in
+    let f () =
```
hannesm, line 31: trailing whitespace

eio/src/mirage_crypto_rng_eio.ml
```diff
@@ -0,0 +1,81 @@
...
+    Entropy.feed_pools None source f
+  in 
+  periodic env task delta
+
+let rdrand_task env delta = 
+  match Entropy.cpu_rng with
+  | Error `Not_supported -> []
+  | Ok cpu_rng -> [ fun () -> periodic env (cpu_rng None) delta ]
+
+let running = ref false
```
hannesm, line 44: trailing whitespace

hannesm:
please remove the trailing whitespaces all over (I started to comment on them, but there are too many).

eio/src/mirage_crypto_rng_eio.mli
hannesm: Done in 96d7ce6bab47584d394d720e18f0366bfa6c343d, does that make sense in that way?

hannesm:
I took the liberty to (a) remove the trailing whitespaces (b) unset the default generator when `run` finishes executing (c) merge the main branch -- which I pushed to your branch @bikallem.

What is left to do:
- minor work on the docstring inconsistency with implementation
- documentation in Mirage_crypto_rng.mli (and rng.ml) to talk about eio (and its special semantics)
- why is it that "eio" is directly at the top level (with the source in `eio/src`), I'd expect it as `rng/eio`; and the tests (in case they can't for some reason be in `tests`, in `rng/eio/tests`.

eio/src/mirage_crypto_rng_eio.mli
talex5: Do we need `running` at all now, then? We can just look at whether the default generator is set.

> in the experience of RNG initialization, having Mirage_crypto_rng_lwt.initialize outside the Lwt monad allowed us to embed the call in Tls_lwt at toplevel / conduit-lwt as well, which means common users of these libraries do not need to think about RNG.

Doesn't that mean that the user's choice of `sleep` duration or of which generator to use would get overridden by the arguments passed by Tls?

hannesm:
@bikallem uhm, but isn't the `dune build -p <lots-of-packages>` in the CI scripts select exactly the packages to be build, and the `dune` files contain `package` lines to put them into specific packages?

bikallem:
@hannesm I have moved `eio/src` and `eio/tests` to `rng/eio` and `tests` respectively and have updated the ci scripts accordingly. I have also updated the documentation in the noted places accordingly.

bikallem:
@hannesm Are there any more issues to be addressed?

hannesm:
Thanks, I merged this. Release will happen shortly.

.cirrus.yml
hannesm: ok, thanks for your reply. now that it'll soon be merged into opam-repository, would you mind to PR something that removes the mentions of specific versions and packages?

.github/workflows/test.yml
```diff
@@ -21,13 +21,52 @@ jobs:
...
         uses: ocaml/setup-ocaml@v2
         with:
+          opam-local-packages: |
+            *.opam
+            !mirage-crypto-rng-eio.opam
           ocaml-compiler: ${{ matrix.ocaml-version }}
 
       - name: Install dependencies
-        run: opam install --deps-only -t .
+        run: opam install --deps-only -t mirage-crypto mirage-crypto-rng mirage-crypto-rng-mirage mirage-crypto-pk mirage-crypto-ec mirage-crypto-rng-async
```
hannesm, line 30: I've no idea why here there are explicit names for all the packages, when above `opam-local-packages` already should contain exactly this list?

bikallem: I believe this is so that it doesn't pick up mirage-crypto-rng-eio.opam dependencies. I think the workflow failed if this line wasn't there then. Perhaps it has changed now, but not sure.

PR #156, opened 2022-05-31
ec: update generated code with recent fiat-crypto

hannesm:


hannesm:
change is marginal, see https://github.com/mit-plv/fiat-crypto/issues/1164 -- mainly the behaviour changes with clang as msvc-cl

PR #163, opened 2022-09-07
remove mirage3 cross-compilation runes

hannesm:
I think this is fine, but of course testing or looking into this would be nice. It removes the mirage3 compatibility (and generates fewer artifacts).


hannesm:
CI fixes are in #164, the removal is delayed (to the next major release).

PR #173, opened 2023-03-09
Test cases for symmetric ciphers with empty input for encrypt and deccrypt

### LINKED ISSUE #172: Mirage_crypto.Chacha20.authenticate_encrypt seg-faults on empty message
(using a 32 bytes key and an 8 bytes nonce) with this:

```ocaml
authenticate_encrypt ~key ~nonce (Cstruct.of_string "")
```

I get a `SIGSEGV`.

hannesm:
Also, fix #172 by providing a `0` case in the chache20 crypt loop


reynir:
This is as well what I arrived at in https://github.com/reynir/mirage-crypto/tree/fix-chacha20. I tested a few values against a python implementation and got the same results. I can't say I understand chacha20 so take it with a LD50 dose of salt.

PR #174, opened 2023-03-09
mirage-crypto-rng.unix: add Hurd support

pinotree:
Implement `raw_getrandom()` for Hurd, using `getrandom()` as available in GNU libc.


hannesm:
Thanks @pinotree. Is there a path to test such a Hurd system in CI (GitHub actions)? Is the C define of Hurd really `__GNU__`, isn't there something like `__HURD__` as well? (`__GNU__` looks pretty coarse, and other systems may define this as well).

hannesm:
Also, can't we fold that into the `__linux` `ifdef` (starting at line 8), since the actual code is the same.

pinotree:
> Is there a path to test such a Hurd system in CI (GitHub actions)?

Unfortunately not.

> Is the C define of Hurd really `__GNU__`, isn't there something like `__HURD__` as well? (`__GNU__` looks pretty coarse, and other systems may define this as well).

Considering GNU/Hurd used to be considered "the GNU system", it got `__GNU__` as identifier; it has also `__gnu_hurd__`. There is no risk that other OSes will use `__GNU__` now, as it would clash with GNU/Hurd.

[1] https://github.com/cpredef/predef/blob/master/OperatingSystems.md#gnu-aka-gnuhurd

> Also, can't we fold that into the `__linux` `ifdef` (starting at line 8), since the actual code is the same.

Sure. Sometimes sharing code branches with other OSes is helpful, sometimes only adds extra maintenance burder.

hannesm:
Thanks. Also thanks for the link to the `predef`.

PR #175, opened 2023-05-03
Stricter C function prototypes

MisterDA:
Modern compilers emit warnings when using K&R C function prototypes and functions taking `()` instead of void for no parameters.

### DIFF
--- a/rng/unix/mc_getrandom_stubs.c
+++ b/rng/unix/mc_getrandom_stubs.c
@@ -35,8 +35,8 @@ void raw_getrandom (uint8_t *data, uint32_t len) {
 #include <sys/param.h>
 
 void raw_getrandom (uint8_t *data, uint32_t len) {
-  int rlen = 0;
-  for (int i = 0; i <= len; i += 256) {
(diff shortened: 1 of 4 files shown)


hannesm:
thanks, the FreeBSD eio CI failure can be ignored (reported https://github.com/ocaml-multicore/eio/issues/502)

rng/unix/mc_getrandom_stubs.c
```diff
@@ -36,7 +36,7 @@ void raw_getrandom (uint8_t *data, uint32_t len) {
 
 void raw_getrandom (uint8_t *data, uint32_t len) {
   int rlen = 0;
```
avsm, line 38: This could be a size_t as well, given the use below in getentropy

avsm:
Compiling for me on clang 14 on macOS (which hits the warnings without this PR)

rng/unix/mc_getrandom_stubs.c
MisterDA: good catch, thanks

PR #176, opened 2023-05-16
mirage-crypto-rng-eio: use monotonic clock from eio's environment instead of mtime 

TheLortex:
This PR improves mirage-crypto-rng-eio's portability by making it not depend on system features. 

Instead of depending on mtime.clock.os, the user should provide a monotonic clock implementation as part of the `env`ironment variable.

### DIFF
--- a/rng/eio/dune
+++ b/rng/eio/dune
@@ -1,4 +1,4 @@
 (library
  (name mirage_crypto_rng_eio)
  (public_name mirage-crypto-rng-eio)
- (libraries eio cstruct logs mirage-crypto-rng duration mtime.clock.os))
+ (libraries eio cstruct logs mirage-crypto-rng duration))
--- a/rng/eio/dune
+++ b/rng/eio/dune
@@ -1,4 +1,4 @@
 (library
  (name mirage_crypto_rng_eio)
  (public_name mirage-crypto-rng-eio)
- (libraries eio cstruct logs mirage-crypto-rng duration))
+ (libraries eio cstruct logs mirage-crypto-rng duration mtime))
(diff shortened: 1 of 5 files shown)


samoht (approved):

rng/eio/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name mirage_crypto_rng_eio)
  (public_name mirage-crypto-rng-eio)
- (libraries eio cstruct logs mirage-crypto-rng duration mtime.clock.os))
+ (libraries eio cstruct logs mirage-crypto-rng duration))
```
hannesm, line 4: Isn't mtime still used (`Mtime.to_unit64_ns` below)? Would it be worth to retain `mtime` in the dune file (if not, should we remove `mtime` from the mirage-crypto-rng-eio opam file as well)?

hannesm:
I don't know much about eio, but what are the semantics / guarantees of `Eio.Stdenv.mono_clock env |> Eio.Time.Mono.now`?

rng/eio/dune
TheLortex: You're right. `mtime` should be kept in the dune file.

TheLortex:
> what are the semantics / guarantees of Eio.Stdenv.mono_clock env |> Eio.Time.Mono.now?

According to the documentation (https://ocaml.org/p/eio/latest/doc/Eio/Time/Mono/index.html), 
_Monotonic clocks are unaffected by corrections to the real-time clock, and so are a better choice for timeouts or measuring intervals, where the absolute time doesn't matter. A monotonic clock may or may not include time while the computer is suspended._

hannesm:
Thanks a lot for your contribution. I squashed and merged this PR, and it will be part of the next release.

PR #177, opened 2023-05-23
Update fiat

hannesm:


hannesm:
hmm, but since there are no code changes, I doubt it'll work on your systems.

samoht:
I indeed do see the same error as before with this PR.

dmmulroy:
@samoht @hannesm I'm curious what issues you were experiencing on osx? I'm running into some issues ES256 in `ocaml-jose` that I've seemed to narrow down to `Mirage_crypto_ec.P256.Dsa.pub_of_cstruct`

https://github.com/ulrikstrid/ocaml-jose/issues/63

samoht:
@dmmulroy see the discussions here: https://github.com/mit-plv/fiat-crypto/issues/1606 it seems to be due to a bug introduced in clang 14.0.3. I've reported an issue to Apple, but no news so far...

dmmulroy:
@samoht Thank you for the context and link - I'll keep my eyes peeled on the issues 👀

hannesm:
since the main difference is "inline everything", a performance evaluation is needed before a merge.

hannesm:
since it is only inlining the change, I'll defer to a later point.

PR #178, opened 2023-06-04
Build powerpc64 with -Wno-stringop-overflow tests due to false positives

avsm:
They do seem to be false positives when I look through the code, and it is also independently reported here:
https://bugs.launchpad.net/ubuntu/+source/gcc-12/+bug/2013140

The warnings disappear at -O2, but compilation and tests seem fine at -O3 with the warning disabled, so I chose to keep it at -O3.

An example warning is:

```
In file included from native/poly1305-donna.c:11:
native/poly1305-donna.c: In function ‘mc_poly1305_update’:
native/poly1305-donna-64.h:24:23: note: at offset 28 into destination object ‘buffer’ of size 16
   24 |         unsigned char buffer[poly1305_block_size];
      |                       ^~~~~~
In function ‘poly1305_update’,
    inlined from ‘mc_poly1305_update’ at native/poly1305-donna.c:60:3:
native/poly1305-donna.c:48:54: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
   48 |                         st->buffer[st->leftover + i] = m[i];
      |                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
```

Spotted while testing https://github.com/ocaml/ocaml/pull/12276

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## dev
+
+* Build fix on PowerPC64 due to spurious gcc warnings for stringop overflow.
+
 ## v0.11.1 (2023-03-09)
 
 * BUGFIX Chacha20 decrypt and encrypt with empty data (previously lead to
--- a/config/cfg.ml
+++ b/config/cfg.ml
@@ -7,11 +7,12 @@ let () =
       Configurator.V1.C_define.import
         c
         ~includes:[]
-        [("__x86_64__", Switch); ("__i386__", Switch)]
+        [("__x86_64__", Switch); ("__i386__", Switch); ("__powerpc64__", Switch)]
     in
     match defines with
     | (_, Switch true) :: _ -> `x86_64
     | _ :: (_, Switch true) :: _ -> `x86
+    | _ :: _ :: (_, Switch true) :: _ -> `ppc64
     | _ -> `unknown
   in
   let accelerate_flags =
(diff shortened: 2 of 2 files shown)


hannesm:
I'm uneasy with this change. It is a big hammer for such an issue: now we compile everything on power without `-Wno-stringop-overflow` -- is this the intention? Could we disable the warning more locally? Could we check for the specific compiler version(s) that encounter this "false positive" (is it confirmed by the gcc folks that it is a false positive?)

FWIW, why is it only added to `opt_flags`?

Isn't that just a compiler warning, and `-Werror` is only enabled in dev profile (via `src/dune`)? Shouldn't that the place be to express `-Werror`, but on power do not enable `no-stringop-overflow` as error (but leave as warning)?

avsm:
It's not confirmed by upstream (as far as I can tell, but I haven't finished trawling the gcc bug tracker yet).  I added it to opt_flags since it only strikes when -O3 is active, but changing this to -Wno-error-stringop-overflow would avoid disabling it and also avoid Werroring (but make the build quite spammy).  Perhaps thats better though -- if there is a compiler bug, it should at least be visible.

hannesm:
So indeed, "if there's a compiler bug, it should at least be visible". Do you plan to work on taking a slightly different road (disable the flags only in the specific environment where they're needed to be excluded)?

What about the CI, is there any for powerpc64 that fails? If not, could this be added?

hannesm:
thanks for your work, again. I merged #184, where the no-strinop-overflow is applied in a much smaller context.

PR #181, opened 2023-07-21
random support for dflybsd

movepointsolutions:
I just added support for __DragonFly__ macro meaning the system is DragonFlyBSD, where getrandom() is the same as for OpenBSD


hannesm:
Thanks a lot!

PR #182, opened 2023-08-30
Update to Eio 0.12

talex5:
Eio 0.12 no longer uses OCaml objects for OS resources. This requires a minor change to mirage-crypto-rng-eio to use the new variant types instead.

(tested using my [Eio 0.12 branch of ocaml-tls](https://github.com/talex5/ocaml-tls/tree/eio))

### DIFF
--- a/mirage-crypto-rng-eio.opam
+++ b/mirage-crypto-rng-eio.opam
@@ -15,7 +15,7 @@ build: [ ["dune" "subst"] {dev}
 depends: [
   "ocaml" {>= "5.0.0"}
   "dune" {>= "2.7"}
-  "eio" {>= "0.7"}
+  "eio" {>= "0.12"}
   "cstruct" {>= "6.0.0"}
   "logs"
   "mirage-crypto-rng" {=version}
--- a/rng/eio/mirage_crypto_rng_eio.ml
+++ b/rng/eio/mirage_crypto_rng_eio.ml
@@ -1,11 +1,11 @@
-
 open Mirage_crypto_rng
 
-type env = <
-  clock: Eio.Time.clock;
-  mono_clock: Eio.Time.Mono.t;
-  secure_random: Eio.Flow.source;
->
+type 'a env = <
+  clock: [> ] Eio.Time.clock;
+  mono_clock: [> ] Eio.Time.Mono.t;
+  secure_random: [> ] Eio.Flow.source;
+  ..
+> as 'a
 
 let src = Logs.Src.create "mirage-crypto-rng-eio" ~doc:"Mirage crypto RNG Eio"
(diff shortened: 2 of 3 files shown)


bikallem:
Seems straightforward. LGTM.

haesbaert:
> Seems straightforward. LGTM.
> 
> @haesbaert ocaml-tls depends on this package so would appreciate if we could expedite a new release of mirage-crypto-rng-eio

I'm sure you meant @hannesm here though :)

bikallem:
> I'm sure you meant @hannesm here though :)

Oops, sorry. Yes, I meant @hannesm.

talex5:
There are some CI failures here (`error: writing 1 byte into a region of size 0` and `Cannot parse point: point is not on curve`), but they are unrelated to this PR and are also failing on the main branch.

As this PR is only changing a type, and only affects the Eio package, it should be safe to merge.

hannesm:
Thanks for your PR. Indeed the CI failures look unrelated (see #178 and a lengthy discussion in https://github.com/mit-plv/fiat-crypto/issues/1606#issuecomment-1560122239).

In order to get a new release of mirage-crypto, I'd like to solve these outstanding issues, and merge this PR as well. I hope to find some time next week to get around cutting a release.

Just to be sure we're on the same page -- the `Eio.Flow.read_exact env#secure_random buf;` and `Eio.Time.sleep env#clock (Duration.to_f delta);` are fine with the most recent eio (I'm asking since in both cases still objects are used..).

talex5:
> Just to be sure we're on the same page -- the `Eio.Flow.read_exact env#secure_random buf`; and `Eio.Time.sleep env#clock (Duration.to_f delta)`; are fine with the most recent eio (I'm asking since in both cases still objects are used..).

Yes, that's fine (`clock` and `secure_random` are no longer objects, but `env` is).

We could pass the resources to mirage-crypto as separate arguments instead of bundling them up in an object, e.g.

```diff
 val run
   :  ?g:'a
   -> ?sleep:int64
   -> 'a Mirage_crypto_rng.generator
-  -> _ env
+  -> clock:_ Eio.Time.clock
+  -> mono_clock:_ Eio.Time.Mono.t
+  -> secure_random:_ Eio.Flow.source
   -> (unit -> 'b) -> 'b
```

However, continuing to use an object here means code using mirage-crypto doesn't need to change, and it allows adding extra resources without requiring changes to user code in the common case (e.g. I see that #176 added a monotonic clock that way).

BTW, the only use of the non-monotonic clock is to sleep for 10s, which could also be done with the monotonic clock, avoiding the need for `clock` at all. If you'd like `clock` removed, I could do that here, or in a separate PR.

PR #137, opened 2021-08-26
Support Microsoft CL.EXE compiler

jonahbeckford:
* `__attribute__((unused))`: Convert to MSVC non-standard extensions rather than GCC non-standard extensions
* Use `_WIN64` in addition to `__x86_64__` to detect 64 bit
* Quote shell parameters
* Drop to 32-bit for some C code on 64-bit Windows to avoid algorithms that use 128-bit integers. MSVC does not support 128 bit integers
* Use MSVC CPU intrinsics rather than GCC intrinsics

PR #183, opened 2023-09-03
Add MSVC to CI

jonahbeckford:
These are the commits after following https://github.com/diskuv/dkml-workflows#configure-your-project

I've set the GitHub jobs to `continue-on-error: true` because MSVC does not work with mirage-crypto. So the jobs won't fail, but at least you can inspect the logs to see what is breaking MSVC.

This goes along with the old https://github.com/mirage/mirage-crypto/pull/137, although at this point I don't know what else is a problem for mirage-crypto with Windows or MSVC.


hannesm:
thanks for your PR. I'm a bit puzzled to include so much code and binaries into the repository, and wonder what the maintenance story behind that should be? Or do the binaries never need to be updated?

jonahbeckford:
Now I'm confused. When you say "... *binaries* into the repository", what binaries are you referring to? Are you referring to the shell scripts?

If you are referring to shell scripts, the maintenance story is:
1. `./dk dkml.wrapper.upgrade` will upgrade the scripts `./dk`, `./dk.cmd` and `cmake/FindDkToolScripts.cmake`. You should never need to do that unless I make a major version upgrade or a security announcement.
2. The other 4 scripts can be recreated with `./dk dkml.workflow.compilers CI GitHub Desktop OS Windows`, but only if the scripts are erased first. I anticipate you would only want to do that with `.github/workflows/dkml.yml`, so those same instructions are explicitly listed inside that file.

Finally ...

The bulk of the GitHub logic is downloaded on-demand, so that is always up-to-date (confer: `run: ./dk dkml.workflow.compilers CI GitHub` in `.github/workflows/dkml.yml`). You don't need to do it on-demand ... you can check those generated files into source control (effectively pinning the GitHub logic).

Similarly, the two `.ci/dkml-compilers/pc` scripts do not need to be committed ... you can tell anyone who is developing/testing on their Windows desktops to first run `./dk dkml.workflow.compilers CI Desktop OS Windows`. That would have the side-benefit of always being up-to-date, but of course it would be less repeatable for fellow developers.

The "HELP" command listed in `.github/workflows/dkml.yml` is a good reference as well.

hannesm:
I was refering to BIN +220 KB
[.ci/dkml-compilers/pc/setup-dkml-windows_x86.ps1](https://github.com/mirage/mirage-crypto/pull/183/files#diff-40372176f6208381689062af78d65f75ecede27bd746ff1bf1683f55c7cbe52f)

and BIN +220 KB
[.ci/dkml-compilers/pc/setup-dkml-windows_x86_64.ps1](https://github.com/mirage/mirage-crypto/pull/183/files#diff-3578f4b43e6150bc59ab2732a65d199fd73d56ba39d6fe07ac75e6bbbd6f0f82)

But, as you say, they are text files... :)

Thanks a lot, could you rebase onto the main branch, I'll rerun the CI, and merge?

jonahbeckford:
Interesting. I'll change the .gitattributes to see if I can stop GitHub from thinking they are binary files. Then I'll rebase

hannesm:
thanks a lot

PR #184, opened 2023-09-18
avoid stringop-overflow warning on ppc64, s390x

hannesm:
superseeds #178


hannesm:
merging, the CI (OCaml-CI) on ppc64 and s390x are passing (and that's the only thing that changed here).

PR #189, opened 2024-01-27
Add NetBSD to the list of BSD OS in define

drchrispinnock:
I currently working on getting Octez (Tezos) to work on NetBSD. Today I successfully built mirage-crypto-rng with this change. I have not been able to do much testing.


hannesm:
Thanks. Do you have any (free to use) CI system at hand that uses NetBSD (to avoid issues in the future)?

drchrispinnock:
Apologies - I do not.

hannesm:
Thanks.

PR #190, opened 2024-01-30
Add support for Loongarch

fangyaling:


hannesm:
any chance there's a CI system that can check these conditional code paths, to avoid regressions in the future?

fangyaling:
It would be good if we could add something to the CI, but now the ocaml patch not merged, only built in local.
Now the code in the patch is refer kernel module.
The ref code link in kernel is  https://lore.kernel.org/lkml/20220226110338.77547-20-chenhuacai@loongson.cn/
+static __always_inline u64 __arch_get_hw_counter(s32 clock_mode,
+						 const struct vdso_data *vd)
+{
+	unsigned int count;
+
+	__asm__ __volatile__(
+	"	rdtime.d %0, $zero\n"
+	: "=r" (count));
+
+	return count;
+}

src/native/entropy_cpu_stubs.c
```diff
@@ -143,6 +143,14 @@ static inline unsigned long get_count(void) {
 }
 #endif
 
+#if defined (__loongarch__)
```
hannesm, line 146: here it is `__loongarch__`

src/native/mirage_crypto.h
```diff
@@ -47,7 +47,7 @@ extern struct _mc_cpu_features mc_detected_cpu_features;
 
 #endif /* __mc_ACCELERATE__ */
 
-#if defined (__x86_64__) || defined (__aarch64__) || defined (__powerpc64__) || (64 == __riscv_xlen) || defined (__s390x__) || (defined (__mips__) && _MIPS_SIM==_ABI64)
+#if defined (__x86_64__) || defined (__aarch64__) || defined (__powerpc64__) || (64 == __riscv_xlen) || defined (__s390x__) || (defined (__mips__) && _MIPS_SIM==_ABI64) || defined(__loongarch64)
```
hannesm, line 50: and here `__loongarch64` (thus no trailing `__`).

would you mind to elaborate which defines are used and why? are all `__loongarch__` systems 64 bit? should the code above in `entropy_cpu_stubs.c` be guarded by the same define?

loongson-zn (approved):

fangyaling:
@hannesm Thanks for your review. And We use __loongarch_lp64 and __loongarch64 (no trailing ) for this case. The __loongarch_lp64 is recommended.

hannesm:
Thanks.

PR #192, opened 2024-02-06
add EC to bench/speed

hannesm:
this adds sign and verify operations of the nist curves to bench/speed.ml


hannesm:
there were sufficiently many CI systems successful with this change.

PR #194, opened 2024-02-16
riscv64: Replace rdcycle64 with rdtime64

AdrianBunk:
rdcycle is privileged since kernel 6.6, resulting in SIGILL:
https://buildd.debian.org/status/fetch.php?pkg=ocaml-mirage-crypto&arch=riscv64&ver=0.11.2-1%2Bb3&stamp=1708088363&raw=0

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -118,10 +118,15 @@ static inline uint64_t read_cycle_counter(void)
 #endif
 
 #if defined (__riscv) && (64 == __riscv_xlen)
-static inline uint64_t rdtime64(void)
+//since rdcycle is a privileged instruction since linux 6.6, we use rdtime when in user-space
+static inline uint64_t cycle_count(void)
 {
   uint64_t rval;
+#if defined(__ocaml_freestanding__) || defined(__ocaml_solo5__)
+  __asm__ __volatile__ ("rdcycle %0" : "=r" (rval));
+#else
   __asm__ __volatile__ ("rdtime %0" : "=r" (rval));
+#endif /* __ocaml_freestanding__ || __ocaml_solo5__ */
   return rval;
 }
 #endif
@@ -160,7 +165,7 @@ CAMLprim value mc_cycle_counter (value __unused(unit)) {
 #elif defined(__powerpc64__)
   return Val_long (read_cycle_counter ());
 #elif defined(__riscv) && (64 == __riscv_xlen)
-  return Val_long (rdtime64 ());
+  return Val_long (cycle_count ());
 #elif defined (__s390x__)
   return Val_long (getticks ());
 #elif defined(__mips__)
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -118,10 +118,10 @@ static inline uint64_t read_cycle_counter(void)
 #endif
 
 #if defined (__riscv) && (64 == __riscv_xlen)
-static inline uint64_t rdcycle64(void)
+static inline uint64_t rdtime64(void)
 {
   uint64_t rval;
-  __asm__ __volatile__ ("rdcycle %0" : "=r" (rval));
+  __asm__ __volatile__ ("rdtime %0" : "=r" (rval));
   return rval;
 }
 #endif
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks for your PR - @edwintorok since you contributed the original code, would approve this change?

edwintorok:
The standard doesn't specify the accuracy of `rdtime`, and since access to `rdcycle` was restricted for "security" reasons it'll likely be lower resolution than `rdcycle`.
I haven't measured myself, but according to https://zhangruiyi.me/files/riscv_attacks_sp23.pdf `rdcycle` gives you 1ns accuracy, whereas `rdtime` gives you 45ns.

So if we replace `rdcycle` by `rdtime` I think we no longer meet the criteria for `mc_cycle_counter` which says that every call should lead to a different output, we'll probably have to maintain our own counter and add that to the value read by `rdtime` to ensure uniqueness.

edwintorok:
It might be better to use `rdcycle` in unikernel mode (where I assume you'd have access to this instruction because you run at a sufficient privilege level), and `rdtime` in userspace mode (where it is better to rely on `/dev/urandom` anyway, very few RISC-V CPUs are out-of-order yet, and on an in-order CPU I doubt the whirlwind loop would introduce much entropy beyond the value of the counter itself, but IIUC the cycle counter is used just as an additional source of entropy,and not the only source, so that is probably fine).

Perhaps use `#if defined(__ocaml_freestanding__) || defined(__ocaml_solo5__)` and use `rdcycle` there as before, and `rdtime` otherwise?

src/native/entropy_cpu_stubs.c
```diff
@@ -118,10 +118,10 @@ static inline uint64_t read_cycle_counter(void)
 #endif
 
 #if defined (__riscv) && (64 == __riscv_xlen)
-static inline uint64_t rdcycle64(void)
+static inline uint64_t rdtime64(void)
```
hannesm, line 121:
```suggestion
//since rdcycle is a privileged instruction since linux 6.6, we use rdtime when in user-space
static inline uint64_t cycle_count(void)
```

src/native/entropy_cpu_stubs.c
```diff
@@ -118,10 +118,10 @@ static inline uint64_t read_cycle_counter(void)
...
 {
   uint64_t rval;
-  __asm__ __volatile__ ("rdcycle %0" : "=r" (rval));
+  __asm__ __volatile__ ("rdtime %0" : "=r" (rval));
```
hannesm, line 124:
```suggestion
#if defined(__ocaml_freestanding__) || defined(__ocaml_solo5__)
  __asm__ __volatile__ ("rdcycle %0" : "=r" (rval));
#else
  __asm__ __volatile__ ("rdtime %0" : "=r" (rval));
#endif /* __ocaml_freestanding__ || __ocaml_solo5__ */
```

src/native/entropy_cpu_stubs.c
```diff
@@ -160,7 +160,7 @@ CAMLprim value mc_cycle_counter (value __unused(unit)) {
 #elif defined(__powerpc64__)
   return Val_long (read_cycle_counter ());
 #elif defined(__riscv) && (64 == __riscv_xlen)
-  return Val_long (rdcycle64 ());
+  return Val_long (rdtime64 ());
```
hannesm, line 163:
```suggestion
  return Val_long (cycle_count ());
```

hannesm:
Thanks @edwintorok -- what Edwin said makes a lot of sense to me, I added code comments that implement his suggestion. Would be great if they could be integrated and tested (I suspect the rdcycle code path is at the moment dead code, since I haven't seen a MirageOS unikernel on RISC-V (yet)).

hannesm:
Could you @AdrianBunk (or @edwintorok) -- or someone else with access to RISC-V hardware please embed the suggestions above and report whether this compiles fine (or not)?

hannesm:
another friendly ping, we're close to a release of mirage-crypto, and it would be great to include this... but without any testing on risc-v (or an approval by e.g. @edwintorok with the suggested changes above), I don't feel comfortable to merge and release.

edwintorok:
I can attempt to do a test this weekend on at least one of my RISC-V boards.

edwintorok (approved):
Tested v0.11.2-20-gdcc8e7c on Sipeed Lichee Pi 4A (rv64imafdcvsu), kernel 5.10.113+, OCaml 5.1.1, gcc 13.2.0:

`dune runtest` worked (except for `async` which is not installable).

PR #195, opened 2024-02-17
mirage-crypto-ec: warn about power/timing analysis on k in sign

hannesm:
//cc @edwintorok since you mentioned this in https://discuss.systems/@edwintorok/111925959867297453 - would this warning suite you?

I'm still unsure what advise to give to be honest: on the one side, it looks that these attacks are local -- thus in a multi-tenant setup, just don't do crypto ;) -- on the other side, asking a developer to select a `k` is something they can easily get wrong as well.

So, below the line: should we switch by default to a random `k`? Should we warn about multi-tenant systems explicitly?

The mitigation you mention "validating that the ciphertext (public key) consists of a pair of linearly independent points of the correct order 3e" is obviously something we can implement (enabled by default) -- maybe measure the performance impact?

Also, you mention blinding, and I'm curious what is in mind -- in our (normal, non-EC) DSA sign_z code we have:
```OCaml
  (* normal DSA sign is: s = k^-1 * (z + r * x) mod q *)
  (* we apply blinding where possible and compute:
     s = k^-1 * b^-1 * (b * z + b * r * x) mod q
     see https://github.com/openssl/openssl/pull/6524 for further details *)
```

Should we do the same in ECDSA sign?

### DIFF
--- a/ec/mirage_crypto_ec.mli
+++ b/ec/mirage_crypto_ec.mli
@@ -112,6 +112,12 @@ module type Dsa = sig
       deterministic construction from RFC 6979. The result is a pair of [r]
       and [s].
 
+      Warning: there {{:https://www.hertzbleed.com/2h2b.pdf}are}
+      {{:https://www.hertzbleed.com/hertzbleed.pdf}attacks} that recover the
+      private key from a power and timing analysis of the RFC 6979 computation
+      of [k] - thus it is advised to provide a good nonce ([k]) explicitly,
+      which is independent of key and digest.
+
       @raise Invalid_argument if [k] is not suitable or not in range.
       @raise Message_too_long if the bit size of [msg] exceeds the curve. *)
 


hannesm:
If I understand correctly, the "linearly independent points" only affects `verify` (i.e. the ciphertext is verified) -- but verify does not involve the private key at all, so I don't quite understand the reasoning behind that.

edwintorok:
> in a multi-tenant setup, just don't do crypto ;) --

I'd agree with you that on current hardware I don't know how to do this securely.  OTOH using some crypto is still better than not using any, so you are at least protected against remote and offline attacks.

There is some hope in hardware-software co-design (see https://trustworthy.systems/projects/timeprotection/) for future RISC-V CPUs.

> would this warning suite you

Thanks, look OK to me.

> select a `k` is something they can easily get wrong as well.

Agreed, I usually prefer if a library/application provides secure defaults, but it is unclear what that should be in this case.
To protect against this timing side-channels it'd be nice if by default it'd use a random `k` instead of deterministic one, but have a flag to provide a custom `k` or ask for the deterministic `k`.
OTOH RFC6969 correctly warns that "Even slight biases in that process may be turned into attacks on the signature schemes.", so although using a random `k` may be useful to defeat timing side-channels, it may expose you to other attacks (I assume that those biases in randomness of `k` may expose you to some offline attacks on signatures created by the same host/program).

Lets go with the warning for now, so at least the user is aware.
The best recommendation I'd have is to use 2 signatures based on 2 different crypto technologies, and require both signatures to be valid (e.g. RSA and ECDSA, although there are many ways to get the RSA signature wrong too). That way even if there is a vulnerability in the implementation (or timing) of one approach you still have to defeat both in order to bypass it.

PR #198, opened 2024-02-22
windows/dkml: opam install with deps-only and with-test only packages to test, deps-only for only build

hannesm:
//cc @jonahbeckford

### DIFF
--- a/.github/workflows/dkml.yml
+++ b/.github/workflows/dkml.yml
@@ -50,10 +50,6 @@ jobs:
     runs-on: ${{ matrix.gh_os }}
     name: build / ${{ matrix.abi_pattern }}
 
-    # mirage-crypto does not support MSVC as of 2023-09-03. Remove
-    # this when MSVC is properly supported.
-    continue-on-error: true
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/ci/build-test.sh
+++ b/ci/build-test.sh
@@ -89,6 +89,7 @@ case "$dkml_host_abi" in
         packages_TOPOLOGICALSORT="$packages_TOPOLOGICALSORT,mirage-crypto-pk,mirage-crypto-ec"
 esac
 #   shellcheck disable=SC2086
+opamrun pin add -y -n .
 opamrun install --yes --deps-only --with-test $(echo $packages_TOPOLOGICALSORT | tr ',' ' ')
 opamrun exec -- dune build -p "$packages_TOPOLOGICALSORT"
 opamrun exec -- dune runtest -p "$packages_TOPOLOGICALSORT"
--- a/ci/build-test.sh
+++ b/ci/build-test.sh
@@ -80,10 +80,11 @@ opamrun update
 
 # Build logic
 #   2024-02-09: Remove mirage-crypto-pk on Windows since no portable GMP library (used by Zarith).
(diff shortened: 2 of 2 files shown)


hannesm:
the std_cflags are properly picked up (in contrast to https://github.com/mirage/mirage-crypto/pull/137#issuecomment-1959389844), also this fixes https://github.com/mirage/mirage-crypto/pull/137#issuecomment-1959395191

remaining is the following issue, on windows 64 bit systems, the fiat 32 bit versions are used (reason behind this is that 128bit unsigned integers are not available) -- now with our pre-computed tables, this results in an error. I guess on a windows 64 bit system, using the 32 bit tables is fine. any other opinion @Firobe @jonahbeckford ?  I added that in [975a881](https://github.com/mirage/mirage-crypto/pull/198/commits/975a881df4b71caf489b716de210b1161a936c9f) (but unfortunately our mirage-crypto-ec tests aren't running on dkml/windows since mirage-crypto-pk and asn1-combinators don't work there (gmp/zarith).

hannesm:
Thanks to #200, we can now run mirage-crypto-ec tests. Unfortunately, they fail:
```
File "tests/dune", line 60, characters 7-25:
60 |  (name test_ec_wycheproof)
            ^^^^^^^^^^^^^^^^^^
(cd _build/default/tests && ./test_ec_wycheproof.exe)
Fatal error: exception File "tests/test_ec_wycheproof.ml", line 418, characters 4-10: Assertion failed
```

It is not entirely clear to me, why, though. The assetion in question:
```OCaml
let to_ed25519_keys (key : eddsa_key) =
  let priv_cs = Cstruct.of_string key.sk
  and pub_cs = Cstruct.of_string key.pk
  in
  match Ed25519.priv_of_cstruct priv_cs, Ed25519.pub_of_cstruct pub_cs with
  | Ok priv, Ok pub ->
    assert (Cstruct.equal Ed25519.(pub_to_cstruct (pub_of_priv priv)) pub_cs); (* that's the failing one *)
    priv, pub
  | _ -> assert false
```

This means that the `(pub_to_cstruct (pub_of_priv _))` isn't equal to the public key. Looks like something is going wrong down the line, but what?

hannesm:
to be clear, I don't have any further energy and time to look deeper into "CL.EXE" and what goes wrong. If someone can pick this up, best with an actual setup to have quick compile-test-edit cycles, that'd be amazing.

I propose to do some printf debugging (looking where `pub_cs` is different from `pub_to_cstruct (pub_to_priv priv)`. It may have to do that OCaml assumes 64bit, and the C code 32bit. There are also a huge amount of warnings from the C compiler, which may be worth to look into.

If there's no interest/time by others, I plan to revert the DKML changes in the main branch to cut a mirage-crypto release.

jonahbeckford:
I have no time to do this at the moment. While I think it would be better just to mark the `ec` opam release as unavailable on windows, do what you think is best.

hannesm:
ok, I removed mirage-crypto-ec from the windows-dkml-ci script. plan to merge this when CI is green.

I have no idea @jonahbeckford about your plans to use mirage-crypto on dkml/windows -- but I'm sure your suggestion "remove mirage-crypto-ec from the build" fits your needs. I'll open an issue about the mirage-crypto-ec and dkml.

PR #203, opened 2024-02-25
Chacha20-Poly1305: use string instead of cstruct

hannesm:
Performance improvement around 2.5x

Worth to gather some more statistics hereof, and the question how to move forward - I can see two ways:
- preserve old API, add some functions (adds noise)
- first reimplement stuff without Cstruct, preserve old API, then cut a release where the API doesn't include cstruct

WDYT?

main branch (28f8cde5ff3197e6383a935037725ab34ab32485)
       16:  11.686102 MB/s  (2318663 iters in 3.028 s)
       64:  44.914143 MB/s  (2248341 iters in 3.055 s)
      256:  111.408313 MB/s  (1464243 iters in 3.209 s)
     1024:  190.362636 MB/s  (530561 iters in 2.722 s)
     8192:  235.326363 MB/s  (81843 iters in 2.717 s)

this PR (with using the ad-hoc API enc_auth_str):
       16:  26.288283 MB/s  (5247580 iters in 3.046 s)
       64:  99.922308 MB/s  (4995212 iters in 3.051 s)
      256:  204.388046 MB/s  (2405283 iters in 2.873 s)
     1024:  278.060783 MB/s  (863503 iters in 3.033 s)
     8192:  294.746967 MB/s  (113552 iters in 3.010 s)

### DIFF
--- a/src/chacha20.ml
+++ b/src/chacha20.ml
@@ -36,36 +36,41 @@ let init ctr ~key ~nonce =
                         bytes (counter 64 bit)."
   in
   let state = Bytes.create block in
-  Bytes.blit_string s 0 state 0 16 ;
-  Bytes.blit_string key 0 state 16 32 ;
+  Bytes.unsafe_blit_string s 0 state 0 16 ;
+  Bytes.unsafe_blit_string key 0 state 16 32 ;
   init_ctr state ;
-  Bytes.blit_string nonce 0 state nonce_off (String.length nonce) ;
+  Bytes.unsafe_blit_string nonce 0 state nonce_off (String.length nonce) ;
   state, inc
 
 let crypt ~key ~nonce ?(ctr = 0L) data =
   let state, inc = init ctr ~key ~nonce in
   let l = String.length data in
   let block_count = l // block in
-  let len = block * block_count in
   let last_len =
     let last = l mod block in
     if last = 0 then block else last
   in
-  let key_stream = Bytes.create len in
+  let res = Bytes.create l in
   let rec loop i = function
     | 0 -> ()
     | 1 ->
-      chacha20_block state i key_stream ;
-      Native.xor_into_bytes data i key_stream i last_len
+      if last_len = block then begin
+        chacha20_block state i res ;
+        Native.xor_into_bytes data i res i block
+      end else begin
+        let buf = Bytes.create block in
+        chacha20_block state 0 buf ;
+        Native.xor_into_bytes data i buf 0 last_len ;
+        Bytes.unsafe_blit buf 0 res i last_len
+      end
     | n ->
-      chacha20_block state i key_stream ;
-      Native.xor_into_bytes data i key_stream i block ;
+      chacha20_block state i res ;
+      Native.xor_into_bytes data i res i block ;
       inc state;
       loop (i + block) (n - 1)
   in
   loop 0 block_count ;
-  let res = Bytes.unsafe_to_string key_stream in
-  if l <> len then String.sub res 0 l else res
+  Bytes.unsafe_to_string res
 
 module P = Poly1305.It
 
--- a/src/chacha20.ml
+++ b/src/chacha20.ml
@@ -118,10 +118,4 @@ let authenticate_decrypt ~key ~nonce ?adata data =
     let cipher, tag = Cstruct.split data (Cstruct.length data - P.mac_size) in
     authenticate_decrypt_tag ~key ~nonce ?adata ~tag cipher
 
-let auth_enc_str ~key ~nonce ?(adata = "") data =
-  let poly1305_key = generate_poly1305_key ~key ~nonce in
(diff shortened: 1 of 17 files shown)


hannesm:
note this could be split into pieces, such as introducing the xor_into_bytes function, then moving poly1305 to string, then adding `macl`, then moving chacha to string. on a separate note, in C we directly cast to `uint32_t *` to avoid shifting and reading byte-wise (this could also be done in a separate commit).

palainp:
Neat! Thanks for that!
Here are my results (I confirm the x2 improvement on smallest sizes, and improvement everywhere!):
- head
       16:  10.119168 MB/s  (1347821 iters in 2.032 s)
       64:  38.610391 MB/s  (1286782 iters in 2.034 s)
      256:  94.388170 MB/s  (778949 iters in 2.015 s)
     1024:  144.286801 MB/s  (296597 iters in 2.007 s)
     8192:  179.547401 MB/s  (45952 iters in 1.999 s)
- this pr
       16:  20.138118 MB/s  (4008503 iters in 3.037 s)
       64:  74.277972 MB/s  (3603763 iters in 2.961 s)
      256:  142.762819 MB/s  (1758954 iters in 3.008 s)
     1024:  182.994248 MB/s  (562805 iters in 3.003 s)
     8192:  189.376331 MB/s  (72417 iters in 2.987 s)

src/chacha20.ml
```diff
@@ -29,81 +29,82 @@ let init ctr ~key ~nonce =
...
     | _ -> invalid_arg "Valid parameters are nonce 12 bytes and key 32 bytes \
                         (counter 32 bit), or nonce 8 byte and key 16 or 32 \
                         bytes (counter 64 bit)."
   in
-  let state = Cstruct.create block in
-  Cstruct.blit_from_string s 0 state 0 16 ;
-  Cstruct.blit key 0 state 16 32 ;
+  let state = Bytes.create block in
+  Bytes.blit_string s 0 state 0 16 ;
+  Bytes.blit_string key 0 state 16 32 ;
```
reynir, line 40: Here we could safely use `Bytes.unsafe_blit_string`.

src/chacha20.ml
```diff
@@ -29,81 +29,82 @@ let init ctr ~key ~nonce =
...
       chacha20_block state i key_stream ;
-      Native.xor_into data.buffer (data.off + i) key_stream.buffer i block ;
+      Native.xor_into_bytes data i key_stream i block ;
       inc state;
       loop (i + block) (n - 1)
   in
   loop 0 block_count ;
-  Cstruct.sub key_stream 0 l
+  let res = Bytes.unsafe_to_string key_stream in
+  if l <> len then String.sub res 0 l else res
```
reynir, line 68: We can use a temporary buffer for the last block if `data` is not aligned with the block size and avoid the last copy.

```patch
-  let key_stream = Bytes.create len in
+  let res = Bytes.create l in
   let rec loop i = function
     | 0 -> ()
     | 1 ->
-      chacha20_block state i key_stream ;
-      Native.xor_into_bytes data i key_stream i last_len
+      if last_len = block then begin
+        chacha20_block state i res ;
+        Native.xor_into_bytes data i res i block
+      end else begin
+        let buf = Bytes.create block in
+        chacha20_block state 0 buf ;
+        Native.xor_into_bytes data i buf 0 last_len ;
+        Bytes.unsafe_blit buf 0 res i last_len
+      end
     | n ->
-      chacha20_block state i key_stream ;
-      Native.xor_into_bytes data i key_stream i block ;
+      chacha20_block state i res ;
+      Native.xor_into_bytes data i res i block ;
       inc state;
       loop (i + block) (n - 1)
   in
   loop 0 block_count ;
-  let res = Bytes.unsafe_to_string key_stream in
-  if l <> len then String.sub res 0 l else res
+  Bytes.unsafe_to_string res
```

reynir (approved):

hannesm:
with @reynir suggested changes 1e1d179
* [chacha20-poly1305]
       16:  27.772872 MB/s  (5554249 iters in 3.052 s)
       64:  105.891640 MB/s  (5228603 iters in 3.014 s)
      256:  211.486396 MB/s  (2608886 iters in 3.012 s)
     1024:  281.579622 MB/s  (886396 iters in 3.074 s)
     8192:  299.032280 MB/s  (116109 iters in 3.033 s)

src/chacha20.ml
hannesm: 1e1d179 -- I guess we may want to look at other places where to use unsafe_X...

src/chacha20.ml
hannesm: great suggestion, 1e1d179

hannesm:
I wondered about the CI failure, and the Bytes.unsafe_blit_string is only available from OCaml 4.09 on (fine with me to bump the lower OCaml bound)

PR #206, opened 2024-02-27
Use xor_into3 in CBC

reynir:
This adds `xor_into3` which takes two input buffers and stores the xor in a destination buffer. Then in CBC this is used so as to not copy the cleartext buffer. I can observe a small speedup on aes-128-cbc of 0-10%. On my laptop it is mostly in the range of 2-5% while on a server with a Intel(R) Xeon(R) CPU E3-1275 v6 CPU with little load I see a more consistent ~8% speedup.

I tried to use this function in chacha20 and in CTR-mode, but it seemed to result in a slowdown.

I am not sure a small speedup is worth adding more code. Maybe `xor_into` could be implemented with `xor_into3` but maybe the compiler can take advantage of knowing one of the sources is the same as the destination(?)

### DIFF
--- a/src/cipher_block.ml
+++ b/src/cipher_block.ml
@@ -172,15 +172,15 @@ module Modes = struct
 
     let encrypt ~key:(key, _) ~iv src =
       bounds_check ~iv src ;
-      let msg = Cs.clone src in
+      let msg = Cstruct.create_unsafe (Cstruct.length src) in
       let dst = msg.buffer in
-      let rec loop iv iv_i dst_i = function
+      let rec loop iv iv_i src_i dst_i = function
         0 -> ()
-      | b -> Native.xor_into iv iv_i dst dst_i block ;
+      | b -> Native.xor_into3 iv iv_i src.buffer src_i dst dst_i block ;
              Core.encrypt ~key ~blocks:1 dst dst_i dst dst_i ;
-             loop dst dst_i (dst_i + block) (b - 1)
+             loop dst dst_i (src_i + block) (dst_i + block) (b - 1)
       in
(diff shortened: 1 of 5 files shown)


src/cipher_block.ml
```diff
@@ -172,15 +172,15 @@ module Modes = struct
...
       bounds_check ~iv src ;
-      let msg = Cs.clone src in
+      let msg = Cstruct.create_unsafe (Cstruct.length src) in
       let dst = msg.buffer in
-      let rec loop iv iv_i dst_i = function
+      let rec loop iv iv_i src_i dst_i = function
         0 -> ()
-      | b -> Native.xor_into iv iv_i dst dst_i block ;
+      | b -> Native.xor_into3 iv iv_i src.buffer src_i dst dst_i block ;
              Core.encrypt ~key ~blocks:1 dst dst_i dst dst_i ;
```
hannesm, line 180: First, thanks for your PR and your experimentation :)

This is an interesting approach -- I thought about 'let's have a encrypt_xor' in one go (so only one call to C), and maybe even have the loop in C (even fewer calls to C) -- but then I looked into chacha, and there the tight loop is a bit more involved (updates the counter state)

so, I'm not 100% sure this is the way to go (and would like to retain uniformity across the ciphers in terms of what is done in C vs what is done in OCaml).

hannesm:
so, what should we do here? is it worth to rebase on top of our string/bytes code?

reynir:
Since this function only gave a small speedup in cbc mode and seemed to slow down ctr mode and chacha20 I'm not so convinced it's worth adding more C code only to use it in some places. Maybe things look different with string/bytes. I am not too motivated to look into it right now so I will close for now.

PR #207, opened 2024-02-27
Add 32 bit ccm test case

reynir:
I expect this test case to fail on 32 bit architecture due to the following snippet:

https://github.com/mirage/mirage-crypto/blob/3ebc0e3e1bb6c471292b9f0c18afa3ce20d5f051/src/ccm.ml#L54-L57

According to the documentation for `lsl` the result is unspecified if the right hand side of `lsl` (in this case 32) is greater than `Sys.int_size`.


reynir:
So it did indeed fail, and now I pushed something which I believe will fix it on 32 bit.

hannesm:
thanks a lot

PR #208, opened 2024-02-28
mirage-crypto-pk: remove s-expression converters and sexplib0 dependency

hannesm:


reynir:
I would be in favor of removing this. However, I found a few users of the sexp converters by searching GitHub. Most notable are jackline and opam-health-check:

https://github.com/hannesm/jackline/blob/a7acd19bd8141b842ac69b05146d9a63e729230d/src/persistency.ml#L203

https://github.com/OnofreTZK/Blockchain4Me/blob/9d0ee58f51725b29bbbf8e65e977427ba7b34d2c/DSAExample/cr.ml#L12

https://github.com/ocurrent/opam-health-check/blob/6aba82fd0cddaddfc966ace83940007284059ac7/server/backend/admin.ml#L21

Those two use the sexp serializers to store keys. Probably for those two cases deserializers could be implemented manually.

hannesm:
@reynir oh, well spotted. I guess some serialisation format is sensible -- but then for private keys PKCS8 (via `X509.Private_key.encode_{der,pem}`) is best used -- it is as well interoperable (for DSA keys that is not (yet?) implemented).

For jackline, I would just embed the priv_of_sexp in that codebase and use (for newly generated OTR keys) PKCS8. In opam-health-check I guess it makes sense to open an issue about it. The Blockchain4Me doesn't look active, and based on the name I won't put much time into it.

hannesm:
Incentive: decrease of binary size of `bench/speed.exe` (not stripped, build in "dune release mode"). main branch (9cb2ebd606d6e79e3fd22ed5bcf81d21992699d4) 7718680 bytes, this branch 7193120 bytes (around 500kB).

PR #209, opened 2024-02-28
mirage-crypto-ec: remove NIST P224 support

hannesm:
according to sherlocode, only used in X509. Also, the ca-certs-nss trust anchors don't include any P224 certificate.


hannesm:
Incentive (apart from -50000 lines of code): binary size of speed.exe (not stripped, built in dune "release mode") of main branch, 9cb2ebd606d6e79e3fd22ed5bcf81d21992699d4: 7718680 bytes, this branch 7256560 bytes -- around 450kB.

PR #210, opened 2024-02-29
mirage-crypto-ec: move API to string (instead of cstruct)

hannesm:


ec/mirage_crypto_ec.mli
```diff
@@ -49,7 +49,7 @@ module type Dh = sig
       than the group order meaning the public key cannot be the point at
       inifinity. *)
 
-  val key_exchange : secret -> Cstruct.t -> (Cstruct.t, error) result
+  val key_exchange : secret -> string -> (string, error) result
```
hannesm, line 52: note it may be useful to be able to pass a `?off` in here (but that may as well be premature optimization) -- eventually you receive in some data stream the public you want to use in `key_exchange`, and with the revised API you always need to copy it out and create a fresh string. I'm rather undecided about that.

hannesm:
Asking @reynir @Firobe @dinosaure @palainp about opinions? I get some (10% - 20%) speedup for EC operations on my laptop with this.

src/uncommon.ml
```diff
@@ -18,6 +18,36 @@ type 'a iter = ('a -> unit) -> unit
...
+        else
+          let c = digit c in
+          match leftover with
+          | None -> chars, Some (c lsl 4)
+          | Some c' -> (c' lor c) :: chars, None)
+      ([], None) s
+  in
+  let chars = List.rev chars in
+  assert (leftover = None);
+  String.init (List.length chars) (fun i -> char_of_int (List.nth chars i))
```
samoht, line 49: `List.nth` doesn't smell very good here. How large to you expect that list to be? I haven't looked in detail, but maybe all of this could be translated into array operations?

hannesm: Thanks for your comment. Indeed, this whole function `of_hex` should move to a place where it is only available for tests. It is not meant to be used outside of tests.

reynir:
On a Intel(R) Core(TM) i7-9700F CPU @ 3.00GHz I don't see any significant speedup (*maybe* <1% speedup). I'm unsure why not. Maybe it's the CPU architecture. On the other hand I don't observe a slowdown!

I'm in favor of this PR at least for the new interface.

hannesm:
it's not 10%, but here:

main branch (61721c270d4cbd3222950dc21632d88eb69be339):
* [ecdsa-sign]
    P256:  7541.426 ops per second (81300 iters in 10.780)
    P384:  3121.375 ops per second (29994 iters in 9.609)
    P521:  443.783 ops per second (4373 iters in 9.854)
    Ed25519:  10687.773 ops per second (103734 iters in 9.706)

* [ecdsa-verify]
    P256:  1267.141 ops per second (14961 iters in 11.807)
    P384:  544.710 ops per second (5587 iters in 10.257)
    P521:  217.304 ops per second (2211 iters in 10.175)
    Ed25519:  7964.725 ops per second (79113 iters in 9.933)

* [ecdh-share]
    P256:  1759.999 ops per second (17793 iters in 10.110)
    P384:  663.540 ops per second (6988 iters in 10.531)
    P521:  238.729 ops per second (2773 iters in 11.616)
    X25519:  12318.587 ops per second (123762 iters in 10.047)


this PR (ff2b4d473d8ab7a46cab37a107d24bd70e2ea298):
* [ecdsa-sign]
    P256:  7600.079 ops per second (79239 iters in 10.426)
    P384:  3198.844 ops per second (31426 iters in 9.824)
    P521:  434.125 ops per second (4339 iters in 9.995)
    Ed25519:  11074.989 ops per second (109170 iters in 9.857)

* [ecdsa-verify]
    P256:  1502.747 ops per second (14318 iters in 9.528)
    P384:  552.742 ops per second (5490 iters in 9.932)
    P521:  215.123 ops per second (2249 iters in 10.454)
    Ed25519:  8069.081 ops per second (85178 iters in 10.556)

* [ecdh-share]
    P256:  1876.072 ops per second (17458 iters in 9.306)
    P384:  694.453 ops per second (6801 iters in 9.793)
    P521:  267.895 ops per second (2816 iters in 10.512)
    X25519:  12567.524 ops per second (142045 iters in 11.303)

so, likely it is noise in the end. but I prefer the interface over the cstruct one ;)

PR #214, opened 2024-03-15
remove cstruct from mirage-crypto

hannesm:
~~that's stlll WIP (compiles, but there are test failures) -- likely there's something wrong when I transcribed from cstruct to bytes/string.~~

also should be reviewed in terms of where stuff is unecessarily copied and could be avoided to be copied

//cc @reynir who may be interested ;)


src/cipher_block.ml
```diff
@@ -317,26 +328,30 @@ module Modes = struct
 
     let (key_sizes, block_size) = C.(key, block)
 
-    let cipher ~key src dst =
-      if src.len < block_size || dst.len < block_size then
-        invalid_arg "src len %d, dst len %d" src.len dst.len;
-      C.encrypt ~key ~blocks:1 src.buffer src.off dst.buffer dst.off
+    let cipher ~key src ~src_off dst ~dst_off =
+      if String.length src < block_size || Bytes.length dst < block_size then
+        invalid_arg "src len %u, dst len %u" (String.length src) (Bytes.length dst);
```
reynir, line 333:
```suggestion
      if String.length src - src_off < block_size || Bytes.length dst - dst_off < block_size then
        invalid_arg "src len %u, dst len %u" (String.length src - src_off) (Bytes.length dst - dst_off);
```

reynir: maybe the error message would be better if the string lengths and offsets are reported separately.

src/cipher_block.ml
```diff
@@ -317,26 +328,30 @@ module Modes = struct
...
-      if Cstruct.length data < tag_size then
+      if String.length data < tag_size then
         None
       else
-        let data, tag = Cstruct.split data (Cstruct.length data - tag_size) in
+        let data, tag =
+          String.sub data 0 (String.length data - tag_size),
+          String.sub data (String.length data - tag_size) tag_size
+        in
         authenticate_decrypt_tag ~key ~nonce ?adata ~tag data
```
reynir, line 355: We copy `tag` only to copy it again in `authenticate_decrypt_tag` :/

src/cipher_stream.ml
```diff
@@ -43,11 +43,11 @@ module ARC4 = struct
           let sj = s.(j) in
           s.(i) <- sj ; s.(j) <- si ;
           let k  = s.((si + sj) land 0xff) in
-          Cstruct.(set_uint8 res n (k lxor get_uint8 cs n));
+          Bytes.(set_uint8 res n (k lxor string_get_uint8 buf n));
```
reynir, line 46:
```suggestion
          Bytes.set_uint8 res n (k lxor string_get_uint8 buf n);
```

src/ccm.ml
```diff
@@ -1,161 +1,171 @@
...
 let pad_block b =
-  let size = Cstruct.length b in
-  Cs.rpad b (size // block_size * block_size) 0
+  let size = Bytes.length b in
+  Bytes.concat Bytes.empty [ b ; Bytes.make (size // block_size * block_size) '\x00' ]
 
-let gen_adata a =
+let pad_block_str b =
+  let size = String.length b in
+  String.concat "" [ b ; String.make (size // block_size * block_size) '\x00' ]
```
reynir, line 44:
```suggestion
  b ^ String.make (size // block_size * block_size) '\x00'
```
Or use `pad_block b |> Bytes.unsafe_to_string`?

src/ccm.ml
```diff
@@ -1,161 +1,171 @@
...
-let gen_ctr_prefix nonce =
-  let n = Cstruct.length nonce in
-  let small_q = 15 - n in
-  let flag = flags 0 0 (small_q - 1) in
-  (flag <+> nonce, succ n, small_q)
+  let to_pad =
+    let leftover = (Bytes.length lbuf + String.length a) mod block_size in
+    block_size - leftover
+  in
+  Bytes.concat Bytes.empty [ hdr ; lbuf ; Bytes.unsafe_of_string a ; Bytes.make to_pad '\x00' ]
```
reynir, line 68: We could avoid intermediate allocations `lbuf` and the padding

src/ccm.ml
```diff
@@ -1,161 +1,171 @@
...
-  let pre, _, q = gen_ctr_prefix nonce in
-  pre <+> encode_len q i
+  let n = String.length nonce in
+  let small_q = 15 - n in
+  let flag_val = flags 0 0 (small_q - 1) in
+  let buf = Bytes.create 16 in
+  Bytes.set_uint8 buf 0 flag_val;
+  Bytes.unsafe_blit_string nonce 0 buf 1 n;
+  encode_len buf ~off:(n + 1) small_q i;
+  buf
```
reynir, line 78: this is quite similar to what `format` does.

reynir (commented):
The changes in ccm.ml are a bit harder to follow and I will need to revisit.

I will suggest some changes.

src/cipher_block.ml
```diff
@@ -209,42 +206,42 @@ module Modes = struct
...
-    let _cs = create_unsafe tagsize
+      Native.GHASH.keyinit cs 0 k;
+      Bytes.unsafe_to_string k
     let hash0 = Bytes.make tagsize '\x00'
-    let digesti ~key i = (* Clobbers `_cs`! *)
+    let digesti ~key i =
       let res = Bytes.copy hash0 in
-      i (fun cs -> Native.GHASH.ghash key res cs.buffer cs.off cs.len);
-      blit_from_bytes res 0 _cs 0 tagsize; _cs
+      i (fun cs -> Native.GHASH.ghash key res cs 0 (String.length cs));
```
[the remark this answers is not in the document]
hannesm: yes, we should if we always use offset = 0 not pass around the offset. the length I'm not too sure about, and prefer a string.length instead of caml_string_length, but that may be worth to benchmark.

why the offset is still there in this PR is that I wanted to finish something that compiles and where the tests are passing before doing such changes (since I initially tried to do the changes with removing offsets in the same PR and failed since I needed to retract).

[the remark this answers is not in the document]
hannesm: having said this, I'm very fine to remove the offset within this PR :) and if we decide in the future that we would like to use different offsets and lengths, we can reiterate. but first make it simple. :)

src/cipher_block.ml
```diff
@@ -209,42 +206,42 @@ module Modes = struct
 
     let stream ~key ~ctr n =
       let blocks = imax 0 n // block_size in
-      let buf    = Native.buffer (blocks * block_size) in
+      let buf = Bytes.create (blocks * block_size) in
       Ctr.unsafe_count_into ctr ~blocks buf 0 ;
-      Core.encrypt ~key ~blocks buf 0 buf 0 ;
-      of_bigarray ~len:n buf
+      Core.encrypt ~key ~blocks (Bytes.unsafe_to_string buf) 0 buf 0 ;
+      String.sub (Bytes.unsafe_to_string buf) 0 n
```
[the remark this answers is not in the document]
hannesm: indeed, I'm not happy about this code

src/cipher_block.ml
```diff
@@ -256,25 +253,38 @@ module Modes = struct
...
-      | 12 -> let (w1, w2) = BE.get_uint64 nonce 0, BE.get_uint32 nonce 8 in
-              (w1, Int64.(shift_left (of_int32 w2) 32 |> add 1L))
-      | _  -> CTR.ctr_of_cstruct @@
-                GHASH.digesti ~key:hkey @@ iter2 nonce (pack64s 0L (bits64 nonce))
+    let pack64s =
+      let _cs = Bytes.create 16 in
+      fun a b ->
+        Bytes.set_int64_be _cs 0 a;
+        Bytes.set_int64_be _cs 8 b;
+        Bytes.unsafe_to_string _cs
```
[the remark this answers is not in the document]
hannesm: maybe we should evaluate this together with #186 in mind -- we have at various places global buffers (that are bad for OCaml 5, and eventually there's not much performance impact). I'd prefer to keep this "remove global buffers" out of this PR.

hannesm:
thanks for the review, @reynir. indeed CCM could get another pass to reduce allocations and remove code duplications.

src/cipher_block.ml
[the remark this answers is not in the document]
reynir: I'm fine deferring this for another PR :)

src/cipher_block.ml
```diff
@@ -317,26 +328,30 @@ module Modes = struct
...
 
     let authenticate_encrypt ~key ~nonce ?adata cs =
       let cdata, ctag = authenticate_encrypt_tag ~key ~nonce ?adata cs in
-      Cstruct.append cdata ctag
+      cdata ^ ctag
 
-    let authenticate_decrypt_tag ~key ~nonce ?(adata = Cstruct.empty) ~tag cs =
+    let authenticate_decrypt_tag ~key ~nonce ?(adata = "") ~tag cs =
+      let tag = Bytes.of_string tag in
       Ccm.decryption_verification ~cipher ~key ~nonce ~maclen:tag_size ~adata ~tag cs
```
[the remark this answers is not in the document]
hannesm: yes, the `tag` was clobbered previously as well. I'll dig a bit deeper.

src/cipher_block.ml
hannesm: see 906d5a0879934186cbc20d4367212abe68b4bcad, not copied in auth_dec_tag

src/cipher_block.ml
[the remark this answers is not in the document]
hannesm: I removed the offset in 4c96e52

src/cipher_block.ml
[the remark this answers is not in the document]
hannesm: done in 35f035bda427cd4a8a79c3a87b2eeb32462d718d

src/ccm.ml
hannesm: addressed in 2aaf9155c343028d73e5325c3dc1e0d43b5ad6df 9575ba23805e7dec870c094bdaa57eab050b9dbe by yourself ;)

src/ccm.ml
hannesm: 6c0140108b60974b2be9d769a2e3a338870a0e95 removes the padding crazyness :)

PR #215, opened 2024-03-18
use digestif 1.2.0 API

hannesm:


hannesm:
//cc @reynir and @dinosaure

pk/rsa.ml
```diff
@@ -274,69 +269,26 @@ module PKCS1 = struct
...
-         Some (`SHA256, sha256)
-       else if is_prefix sha384 buf then
-         Some (`SHA384, sha384)
-       else if is_prefix sha512 buf then
-         Some (`SHA512, sha512)
-       else
-         None)
+    (fun h ->
+       match List.assoc_opt h map with
+       | None -> invalid_arg "unsupported hash (only MD5 and SHA are supported)"
```
reynir, line 284: Considering digestif implements SHA3 too...
```suggestion
       | None -> invalid_arg "unsupported hash (only MD5, SHA1 and SHA2 are supported)"
```

reynir (approved):

pk/rsa.ml
hannesm: Thanks for the suggestion. This morning while riding my bike I figured that we should avoid these exceptions, and provide a more narrow interface. So I went ahead and pushed c19f54e -- let me know what you think of that. :)

pk/mirage_crypto_pk.mli
```diff
@@ -187,15 +187,12 @@ module Rsa : sig
         was produced with the given [key] as per {{!sig_encode}sig_encode}, or
         [None] *)
 
-    type hash = [ `MD5 | `SHA1 | `SHA224 | `SHA256 | `SHA384 | `SHA512 ]
-    (** The type of supported hash algorithms. *)
-
-    val min_key : hash -> bits
+    val min_key : [ `MD5 | `SHA1 | `SHA224 | `SHA256 | `SHA384 | `SHA512 ] -> bits
```
reynir, line 190: This is great! We can be more clear that these are a subset of `Digestif.hash'`:
```suggestion
    val min_key : [< Digestif.hash' > `MD5 `SHA1 `SHA224 `SHA256 `SHA384 `SHA512 ] -> bits
```

reynir (approved):

pk/mirage_crypto_pk.mli
hannesm: oh yes, that's beatiful. I applied it here and on the other locations. 367366b

PR #218, opened 2024-03-19
use a set for entropy sources

hannesm:
not sure whether this is worth it...


hannesm:
The advantage of this PR is to avoid #94 in first principles, and thus merging this is a good idea :)

hannesm:
It does not fix #186, since still there's a global mutable state that is updated potentially concurrently.

PR #225, opened 2024-04-10
mirage-crypto: skip Cipher_block / Cipher_stream module indirection

### LINKED ISSUE #224: Revise API (remove intermediate modules)
Now that Hash is gone, we could remove Cipher_block and Cipher_stream entirely (and rename `Cipher_block.S` to `Block`, `Cipher_stream.S` to `Stream`).

The usage would then modify from `Mirage_crypto.Cipher_block.AES` to `Mirage_crypto.AES`, which would save some vertical space.

As discussed with @reynir

hannesm:
fixes #224

### DIFF
--- a/bench/speed.ml
+++ b/bench/speed.ml
@@ -1,7 +1,5 @@
 open Mirage_crypto
 
-open Cipher_block
-
 module Time = struct
 
   let time ~n f a =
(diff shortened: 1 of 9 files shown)


reynir (approved):
Awesome! Thanks.

PR #226, opened 2024-04-20
Set a new bytes used for elliptic curves computations to '\000'

dinosaure:
`Cstruct.create` does this. If we don't initialize bytes with '\000', `Field_element.zero` can be something else than '\000'. It's a fix for mirleft/ocaml-x509#167.

### DIFF
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -135,7 +135,7 @@ end
 module Make_field_element (P : Parameters) (F : Foreign) : Field_element = struct
   let b_uts b = Bytes.unsafe_to_string b
 
-  let create () = Bytes.make P.fe_length '\000'
+  let create () = Bytes.create P.fe_length
 
   let mul a b =
     let tmp = create () in
@@ -158,7 +158,8 @@ module Make_field_element (P : Parameters) (F : Foreign) : Field_element = struc
     b_uts tmp
 
   let zero =
-    b_uts (create ())
+    let b = Bytes.make P.fe_length '\000' in
+    b_uts b
 
   let one =
     let fe = create () in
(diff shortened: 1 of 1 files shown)


dinosaure:
> To avoid setting to \000 the tmp Bytes in mul, add and sub (if I understand correctly, it write the entire Bytes), wouldn't it better to leave create() as is, and set to \000 in zero?

Yes, I've thought about this minor optimisation. I leave it to @hannesm to choose whether to be paranoid or optimal.

reynir (approved):
This fixes a critical bug. We can always optimize in a separate PR in my opinion.

hannesm:
first of all, thanks for your investigations and the remarks. :)

the question for me is about symmetry -- there's a `create_octets` function in the same functor, should it use as well `Bytes.make`? And there's a functor `Make_Fn` -- here should the `create` and `create_octets` as well zero out the data?

on a related note, is there any measurable performance impact?

and should we stay in sync with OCaml stdlib and have "create" for "allocate", and "make" for "allocate and initialize" (i.e. rename to `make` / `make_octets`)?

hannesm:
Thanks, I looked a bit deeper, and indeed only the `zero` requires a cleared buffer -- so I amended your PR with a commit that ensures that the `zero` will be a buffer of zeroes.

Let me know if you think this is too little? Or if you can test x509 in your 32 bit container with the added commit?

Measurements on my laptop are rather noisy in this area, but across 3 runs the performance difference between the main branch (47751da) and the latest commit (81c492ef) is not present, while with 3333de37 there's a small decrease.

What do you think? (/cc @reynir @dinosaure @palainp)

dinosaure:
LGTM

palainp:
It's ok for me too :)

hannesm:
thanks, merged. the OCaml-CI in ocaml-x509 also reported success :)

Issue #216, opened 2024-03-19
ARM64 cycle count


hannesm:
Since we switched from Cstruct.t to string, there are some failures on arm64 architectures that `read_virtual_count` returns the same value in subsequent calls to it (see tests/test_entropy.ml).

What we use:
```C
#if defined (__arm__)
/*
 * The ideal timing source on ARM are the performance counters, but these are
 * presently masked by Xen.
 * It would work like this:

#if defined (__ARM_ARCH_7A__)
  // Disable counter overflow interrupts.
  __asm__ __volatile__ ("mcr p15, 0, %0, c9, c14, 2" :: "r"(0x8000000f));
  // Program the PMU control register.
  __asm__ __volatile__ ("mcr p15, 0, %0, c9, c12, 0" :: "r"(1 | 16));
  // Enable all counters.
  __asm__ __volatile__ ("mcr p15, 0, %0, c9, c12, 1" :: "r"(0x8000000f));

  // Read:
  unsigned int res;
  __asm__ __volatile__ ("mrc p15, 0, %0, c9, c13, 0": "=r" (res));
*/
#if defined(__ocaml_freestanding__) || defined(__ocaml_solo5__)
static inline uint32_t read_virtual_count (void)
{
  uint32_t c_lo, c_hi;
  __asm__ __volatile__("mrrc p15, 1, %0, %1, c14":"=r"(c_lo), "=r"(c_hi));
  return c_lo;
}
#else
/* see https://github.com/mirage/mirage-crypto/issues/113 and
https://chromium.googlesource.com/external/gperftools/+/master/src/base/cycleclock.h
   The performance counters are only available in kernel mode (or if enabled via
   a kernel module also in user mode). Use clock_gettime as fallback.
 */
#include <time.h>
static inline uint32_t read_virtual_count (void)
{
  uint32_t pmccntr;
  uint32_t pmuseren;
  uint32_t pmcntenset;
  // Read the user mode perf monitor counter access permissions.
  __asm__ __volatile__ ("mrc p15, 0, %0, c9, c14, 0" : "=r" (pmuseren));
  if (pmuseren & 1) {  // Allows reading perfmon counters for user mode code.
    __asm__ __volatile__ ("mrc p15, 0, %0, c9, c12, 1" : "=r" (pmcntenset));
    if (pmcntenset & 0x80000000ul) {  // Is it counting?
[...]

PR #227, opened 2024-04-24
Add an implementation of `mirage-crypto-rng-miou` to initialize the RNG with Miou

dinosaure:
This is my first attempt at implementing `mirage-crypto-rng-miou`. The idea of this initialization is to create a process that will periodically refill the random number generator. I've just tested a program that lasts several hours and I haven't observed any memory leakage: the tasks created to refill the generator are cleaned up properly (every second by default).

The initialization creates a witness to the promise that the user must stop before the end of the application (otherwise Miou raises the `Still_has_children` exception). `Mirage_crypto_rng_miou.kill` just cancels the task and finishes all its sub-tasks.

Finally, I've re-implemented `Fortuna` to be domain-friendly. This is because access may involve changes to the state of the random number generator. We need to ensure that several domains can generate random numbers without conflicts (data-race). A mutex is therefore used to protect exclusive access to the state of the generator. I will verify my implementation with `tsan`.


dinosaure:
I can confirm via TSan that the current implementation `Pfortuna` does not generates data-races and `Fortuna` does.

dinosaure:
I force-pushed the final version which is rechecked with TSan (I found an issue about Miou itself). The documentation was improved and the `rng` task can be run in parallel or concurrently depending on the number of domains available. 

NOTE: On OCaml 5.1, I was unable to compile tests without the `(modes native)`, so I added them to be able to link correctly C stubs. The error is probably related to the bytecode target and, in some ways, `dune` is not available to link correctly C stubs and bytecode executable.

hannesm:
great to see this happening, but I'm rather uncomfortable with the fork of fortuna.ml -- this has a high maintenance cost. any chance to have a single code base (is there noticable performance if we use mutex everywhere?)

dinosaure:
> this has a high maintenance cost. any chance to have a single code base (is there noticable performance if we use mutex everywhere?)

I don't have strong opinion and we probably can just use `Mutex` (instead of `Miou.Mutex`) to be sure that the `Fortuna` engine is domain-safe for any schedulers. But yes, it requires some benchmarks.

hannesm:
> But yes, it requires some benchmarks.

We have `dune build --rel bench/speed.exe && _build/default/bench/speed.exe fortuna` -- which generates random bytes (but there's no entropy collection happening during that phase, so it may be not the right thing).

dinosaure:
The PR is ready to be reviewed and merged :+1:.

bench/dune
```diff
@@ -2,4 +2,4 @@
  (names speed)
  (modules speed)
  (libraries mirage-crypto mirage-crypto-rng mirage-crypto-rng.unix
-   mirage-crypto-pk mirage-crypto-ec))
+   mirage-crypto-pk mirage-crypto-ec mirage-crypto-rng-miou-unix))
```
hannesm, line 5: no, let's please keep bench/speed.exe working on 4.14, and without any IO library.

you're welcome to contribute a miou-bench.exe.

bench/speed.ml
```diff
@@ -404,6 +404,18 @@ let benchmarks = [
...
+  bm "pfortuna" (fun name ->
+    let open Mirage_crypto_rng_miou_unix.Pfortuna in
+    Miou_unix.run ~domains:2 @@ fun () ->
+    let rng = Mirage_crypto_rng_miou_unix.(initialize (module Pfortuna)) in
+    let g = create () in
+    reseed ~g "abcd" ;
+    throughput name (fun buf ->
+        let buf = Bytes.unsafe_of_string buf in
+        generate_into ~g buf ~off:0 (Bytes.length buf));
+    Mirage_crypto_rng_miou_unix.kill rng) ;
```
hannesm, line 417: see above, this is not the file where miou (or lwt or async or eio) stuff will go. also, above you added an empty file "bench/fortuna.ml"..

hannesm:
looks mostly fine, I had some minor comments. thanks for your contribution.

dinosaure:
It seems that `test_entropy.exe` fails on MacOS (and it's unrelated to this PR). Otherwise, I splitted out the benchmark about `pfortuna` and added a comment about why we re-implemented the module.

hannesm:
> It seems that `test_entropy.exe` fails on MacOS (and it's unrelated to this PR).

Indeed, tracked as #216

hannesm:
Thanks, I rebased on main and force-pushed, after merging #232 there was the only conflict in .cirrus.yml. Also, I added a comment to keep fortuna and pfortuna in sync :)

hannesm:
Once CI passes (see notes in https://github.com/mirage/mirage-crypto/pull/232#issuecomment-2156509775), I'll go ahead and squash-merge this PR.

hannesm:
Hm, I've no clue what to do with bench/miou -- marking it optional doesn't lead to the OCaml-CI not building it -- and marking it with `(package mirage-crypto-rng-miou-unix)` results in a dune error since there's no public_name -- but we don't want a public_name since we don't want it to be installed when installing mirage-crypto-rng-miou-unix.

Any pointers how to get out of this situation? Maybe commenting the dune rule out for the time being?

hannesm:
> Maybe commenting the dune rule out for the time being?

This is what I pushed to this PR. CI looks good. I'll squash & merge if @dinosaure is happy with this PR (esp. with my changes).

hannesm:
> NOTE: On OCaml 5.1, I was unable to compile tests without the `(modes native)`, so I added them to be able to link correctly C stubs. The error is probably related to the bytecode target and, in some ways, `dune` is not available to link correctly C stubs and bytecode executable.

Please note, I had this issue as well, the problem is that if you have mirage-crypto installed, it uses in bytecode mode the installed C stubs instead of the locally compiled ones... And I've a deja-vu https://github.com/ocaml/dune/issues/9979

So, let's remove the `(modes native)`. It is a dune issue, but our tests shouldn't be a scapegoat for that.

dinosaure:
I'm ok with your changes :+1:

PR #230, opened 2024-05-07
[ec] Handle K_gen correctly for bitlen mod 8 <> 0

### LINKED ISSUE #105: EC P521 tests from RFC 6979 are not passing
the underlying problem is that our "generate a deterministic k" implementation is implemented in a way that full bytes are retrieved from the RNG (instead of bits).

### LINKED ISSUE #228: P-521 sign not constant-time
While doing performance measurements I noticed, that P-521 signatures are not constant-time, which I assume they should be. P-256 is not affected. I see the same behavior with and without the recent performance improvements for elliptic curves. However, I'm not using the latest mirage-crypto, but 0.10.7. Here the result of a simple test script, always signing the same message but generating a new key every two measurements:

```
Generating new key
requests:  1402 ( 4 conn) time: 4s rps:  350.56 latency: 10.78ms/11.38ms/19.22ms
requests:  1347 ( 4 conn) time: 4s rps:  336.88 latency: 10.79ms/11.84ms/29.74ms
Generating new key
requests:  1583 ( 4 conn) time: 4s rps:  395.94 latency: 8.77ms/10.07ms/16.48ms
requests:  1591 ( 4 conn) time: 4s rps:  397.87 latency: 8.42ms/10.02ms/17.3ms
Generating new key
requests:  1658 ( 4 conn) time: 4s rps:  414.71 latency: 9.02ms/9.62ms/18.06ms
requests:  1667 ( 4 conn) time: 4s rps:  416.97 latency: 8.95ms/9.56ms/19.09ms
Generating new key
requests:   956 ( 4 conn) time: 4s rps:  239.14 latency: 10.52ms/16.68ms/23.01ms
requests:   951 ( 4 conn) time: 4s rps:  237.89 latency: 16.05ms/16.76ms/30.08ms
Generating new key
requests:  1150 ( 4 conn) time: 4s rps:  287.71 latency: 12.86ms/13.86ms/34.07ms
requests:  1192 ( 4 conn) time: 4s rps:  298.21 latency: 9.87ms/13.37ms/20.07ms
Generating new key
requests:  1240 ( 4 conn) time: 4s rps:  310.17 latency: 11.95ms/12.86ms/30.99ms
requests:  1276 ( 4 conn) time: 4s rps:  319.10 latency: 11.97ms/12.5ms/31.4ms
Generating new key
requests:   695 ( 4 conn) time: 4s rps:  173.82 latency: 12.77ms/22.94ms/29.86ms
requests:   703 ( 4 conn) time: 4s rps:  175.90 latency: 11.16ms/22.67ms/32.57ms
Generating new key
requests:  1261 ( 4 conn) time: 4s rps:  315.45 latency: 10.32ms/12.64ms/29.33ms
requests:  1256 ( 4 conn) time: 4s rps:  314.01 latency: 11.76ms/12.7ms/21.3ms
```

@Firobe could reproduce the effect with a modified mirage-crypto benchmark:

```
* [ecdsa-timing-key1]
    :  566.780 ops per second (5712 iters in 10.078)

* [ecdsa-timing-key2]
    :  480.910 ops per second (4830 iters in 10.043)

* [ecdsa-timing-key3]
    :  338.418 ops per second (3396 iters in 10.035)

* [ecdsa-timing-key4]
    :  681.658 ops per second (6811 iters in 9.992)

* [ecdsa-timing-key5]
    :  554.708 ops per second (5564 iters in 10.031)
```

Firobe:
Closes #228 
Closes #105 

During K generation, we need to retrieve `l` bits from RNG (where `l` is 256, 384, 521).

Currently, we take `ceil(l / 8)` whole bytes and call it a day. This works fine when `l` is a multiple of 8, but for `l = 521` this generates 7 bits of garbage, which make it very likely the generated vector is over 2^521. When this happens, we loop and retry until we're lucky, which is a lot of busy work for nothing (and makes the time to generate `k` fluctuate wildly, see #228).

This PR truncates the generated vector to keep only the `l` leftmost bits, as per RFC 6979, which:
- makes the generated `k` correct on the first try most of the time, like the other curves
- makes the time of signature fluctuate way less (see #228)
- makes the formerly disabled test vectors for P521 pass, so we enable them (#105)

Since we have to keep the leftmost bits, this means we have to shift the entire string instead of just truncating the first byte. Comments are welcome if there are better ways to do that than implement here :)

### DIFF
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -51,6 +51,7 @@ module type Dsa = sig
   type priv
   type pub
   val byte_length : int
+  val bit_length : int
   val priv_of_octets : string -> (priv, error) result
   val priv_to_octets : priv -> string
   val pub_of_octets : string -> (pub, error) result
@@ -85,6 +86,7 @@ module type Parameters = sig
   val n : field_element
   val pident: string
   val byte_length : int
+  val bit_length : int
   val fe_length : int
   val first_byte_bits : int option
 end
@@ -600,6 +602,8 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
 
   let byte_length = Param.byte_length
 
+  let bit_length = Param.bit_length
+
   let priv_of_octets= S.of_octets
 
   let priv_to_octets = S.to_octets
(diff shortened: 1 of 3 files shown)


Firobe:
The CI failures seem unrelated

hannesm:
Thanks for your work. I had minor suggestions in https://github.com/Firobe/mirage-crypto/pull/1

hannesm:
FWIW, main branch, ecdsa-sign:
P521:  553.436 ops per second (5212 iters in 9.418)

This PR (including my minor changes):
P521:  1543.730 ops per second (15010 iters in 9.723)

other curves (256, 384) didn't change in ops per second.

Firobe:
Thanks for the review! The changes are merged.

PR #231, opened 2024-06-07
provide ciphers with {de,en}crypt_into functionality

hannesm:
This avoids buffer allocations. There are as well unsafe functions exposed for elevated performance.

Reviews welcome, this is WIP since for now only ECB and ECB modes have this new interface (still missing CTR, GCM, CCM). Will continue work on that. Will superseed #204 (at a future point in time).

PR #232, opened 2024-06-09
minor updates

hannesm:
extracted from #231

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,20 +14,8 @@ freebsd_task:
     - opam env
 
   pin_packages_script:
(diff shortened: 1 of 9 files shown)


hannesm:
The CI status is not very satisfying, in more detail:
- macOS arm64 (and general arm64) test_entropy doesn't always pass (see #216 for further details) - in earlier mirage-crypto-versions this issue wasn't observable, likely related to the move from bigarray to bytes/string
- opensuse and bin_prot doesn't work (see https://github.com/ocaml/opam-repository/issues/26005, there's as well a PR upstream https://github.com/janestreet/bin_prot/pull/29)
- FreeBSD and core_unix doesn't work (see https://github.com/janestreet/core_unix/issues/11)
- the cirrus MirageOS tests are a bit hackish -- since we need mirage-random at version 4, we pin it manually (mirage itself at the moment emits a bound >= 3 & < 4). Once a mirage version is released which is aware that mirage-crypto doesn't use cstruct anymore, we can revert that change.

But there's nothing in this repository we can do about it.

PR #242, opened 2024-08-20
fix CCM, as discovered when porting TLS to string

hannesm:
The issue were missing offsets, both if multiple blocks were encrypted/decrypted and when unaligned data (i.e. not a multiple of block_size) was encrypted and decrypted.

//cc @reynir @dinosaure

### DIFF
--- a/src/ccm.ml
+++ b/src/ccm.ml
@@ -79,11 +79,11 @@ let crypto_core_into ~cipher ~mode ~key ~nonce ~adata src ~src_off dst ~dst_off 
 
   let small_q = 15 - String.length nonce in
   let ctr_flag_val = flags 0 0 (small_q - 1) in
-  let ctrblock i block =
-    Bytes.set_uint8 block 0 ctr_flag_val;
-    Bytes.unsafe_blit_string nonce 0 block 1 (String.length nonce);
-    encode_len block ~off:(String.length nonce + 1) small_q i;
-    cipher ~key (Bytes.unsafe_to_string block) ~src_off:0 block ~dst_off:0
+  let ctrblock i block dst_off =
+    Bytes.set_uint8 block dst_off ctr_flag_val;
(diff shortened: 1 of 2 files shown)


src/ccm.ml
```diff
@@ -113,14 +113,14 @@ let crypto_core_into ~cipher ~mode ~key ~nonce ~adata src ~src_off dst ~dst_off
...
       let buf = Bytes.make block_size '\x00' in
       Bytes.unsafe_blit dst dst_off buf 0 len ;
-      ctrblock ctr buf ;
+      ctrblock ctr buf 0 ;
       Bytes.unsafe_blit buf 0 dst dst_off len ;
       unsafe_xor_into src ~src_off dst ~dst_off len ;
       Bytes.unsafe_blit_string cbcblock cbc_off buf 0 len ;
       Bytes.unsafe_fill buf len (block_size - len) '\x00';
-      cbc (Bytes.unsafe_to_string buf) cbc_off iv 0
+      cbc (Bytes.unsafe_to_string buf) 0 iv 0
```
reynir, line 121: Hmm is it safe to switch `cbc_off` with `0`?

reynir (approved):

src/ccm.ml
reynir: No, it's not the counter stuff, I am confusing myself. But this looks good.

hannesm: `buf` is the locally created 16 byte buffer that holds the last block (which is then partially blit into dst). the `cbc` is for the tag, and here the full `buf` is used -- where previously the `cbcblock` (depending on encrypt or decrypt, the src or dst) has been blit into, followed by clearing the remaining bytes with 0.

hannesm:
CI failures are:
- arm issue #216 (GitHub macOS runners, OCaml-CI macos & linux runners)
- fetch failures on OCaml-CI
- FreeBSD failure on OCaml-CI due to core issue (timerfd)

PR #244, opened 2024-08-20
mirage-crypto-ec: provide Dh.secret_to_octets

### LINKED ISSUE #243: ECDH private key serialization
Dear devs,
First of all thank you for this great library :)
In order to perform some local tests, I've needed to export (as a string) the ECDH priv/pub keys. So far, the library uses a secret that I haven't managed to export due to my lack of internal understanding :(
Is this (serialising the ECDH secret) something that makes sense and that it would be possible to write?
Best.

hannesm:
fixes #243 //cc @palainp


hannesm:
I was too impatient to wait for all the CI systems. The overall result look good. :)

palainp:
Thank you!

PR #248, opened 2024-10-10
Mingw friendly case for includes.

mefyl:
Since windows filesystem is case insensitive and, AFAIK, `windows.h` is always lowercase in MinGW distributions, using a lowercase include will work on both transparently.

Currently:

```
mc_getrandom_stubs.c:60:10: fatal error: Windows.h: No such file or directory
   60 | #include <Windows.h>
      |          ^~~~~~~~~~~
```

### DIFF
--- a/rng/unix/mc_getrandom_stubs.c
+++ b/rng/unix/mc_getrandom_stubs.c
@@ -57,7 +57,7 @@ void raw_getrandom (uint8_t *data, uint32_t len) {
  * have decided to go with the more modern Windows API with bcrypt,
  * and make Windows 10 our minimum supported version of mirage-crypto.
  */
-#include <Windows.h>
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks for your PR. By chance, would you be able to add a GitHub action (or another CI) that tests the "mingw" case?

mefyl:
I'm not really familiar with Github actions; is there already a CI environment cross compiling with mingw ?

hannesm:
> I'm not really familiar with Github actions; is there already a CI environment cross compiling with mingw ?

I don't know. Maybe ask the people who're into "windows" and "OCaml" - they must have some secret runes how to test their projects in CI / GitHub actions.

hannesm:
Thanks.

PR #250, opened 2024-11-27
provide new generators: urandom and getentropy

### LINKED ISSUE #249: mirage-crypto-rng.unix thread safety?
I don't see any warnings about thread safety in `mirage-crypto-rng`, but looking at the implementation:
```
rng/fortuna.ml
27:(* XXX Locking!! *)
```

And there various changes to mutable fields without locks, so I think there is a chance that running `Mirage_crypto_rng.generate` would generate the same random numbers in 2 different threads, which would be quite bad (or at least I can't prove that it wouldn't).

If the library is not meant to be thread safe, please add a warning to the .mli

This may not be a problem for mirage (which uses Lwt, so typically single-OCaml-threaded, and not thread safe), but is a problem for multi-threaded OCaml code that would want to use Mirage_crypto_rng.

hannesm:
to fix #249


hannesm:
There's the question how to annotate Fortuna to be racy. Should there be an `[@@alert]`?

rng/unix/mirage_crypto_rng_unix.ml
```diff
@@ -1,13 +1,32 @@
...
+  | _ -> use_getentropy ()
+
 let src = Logs.Src.create "mirage-crypto-rng.unix" ~doc:"Mirage crypto RNG Unix"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-external getrandom_buf : bytes -> int -> unit = "mc_getrandom" [@@noalloc]
+external getrandom_buf : bytes -> int -> int -> unit = "mc_getrandom" [@@noalloc]
+
+let getrandom_into buf ~off ~len =
+  getrandom_buf buf off len
```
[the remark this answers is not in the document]
hannesm: let's remove it, it was only temporarily for the speed test (we can adapt the benchmark now with the right generators)

rng/unix/urandom.ml
```diff
@@ -0,0 +1,27 @@
+
+type g = In_channel.t * Mutex.t
+
+let block = 2048
+
+let create ?time:_ () =
+  let ic = In_channel.open_bin "/dev/urandom"
```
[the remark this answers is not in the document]
hannesm: likely it is a good idea :)

hannesm:
The failure semantics is now a bit sad: if for some reason the fd that has /dev/urandom open isn't able to read, we'll end up in an exception. previously with fortuna we were sure that once it is seeded, we'll never fail...

but i guess if /dev/urandom doesn't deliver anymore we have other problems, and it's fine to raise an exception.

rng/unix/urandom.ml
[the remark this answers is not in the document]
dinosaure: We probably should also add a `at_exit` to properly close it too at the end of the program.

rng/unix/mirage_crypto_rng_unix.ml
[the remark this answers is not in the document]
hannesm: I removed this in 3abb84a51bf7ead515debcdf889544b2ada688fd

rng/unix/getentropy.ml
```diff
@@ -0,0 +1,20 @@
...
+external getrandom_buf : bytes -> int -> int -> unit = "mc_getrandom" [@@noalloc]
+
+type g = unit
+
+let block = 256
+
+let create ?time:_ () = ()
+
+let generate_into ~g:_ buf ~off len =
+  getrandom_buf buf off len
```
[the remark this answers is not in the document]
hannesm: It is no longer (3abb84a51bf7ead515debcdf889544b2ada688fd) -- also the Mirage_crypto_rng.generate_into does the required checks.

rng/unix/urandom.ml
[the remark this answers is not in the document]
hannesm: I added the `at_exit` thing in 0917c36f35f6cba100a91fcdc256bfafde649412

and also some documentation in 36562df4173afc6781308385baae09aaede7740e

hannesm:
I think this is fine to merge now.

hannesm:
@edwintorok happy new year! It'd be great if you could have a look at this PR -- together with a release and announcement on discuss.ocaml.org, I think this would solve #249. What do you think?

rng/unix/urandom.ml
```diff
@@ -0,0 +1,27 @@
+
+type g = In_channel.t * Mutex.t
+
+let block = 2048
```
edwintorok, line 4: Should this be set to 256?

The Linux manpage says:
>  When calling read(2) for the device
>       /dev/urandom, reads of up to 256 bytes will return as many bytes as are requested and will not be in‐
>       terrupted by a signal handler.  Reads with a buffer over this limit may  return  less  than  the  re‐
>       quested number of bytes or fail with the error EINTR, if interrupted by a signal handler

>  Reads from this device do not block (i.e., the CPU is not yielded),  but  can
>       incur an appreciable delay when requesting large amounts of data

OCaml's input channel will cope with EINTR and short reads, although if we only need a small amount of entropy then reading more than 256 may not be beneficial.
Although a test on my system shows it only returns early when reading more than 2047MiB in one go with `dd`.

Looking at FreeBSD's manual there is no recommended size.

In the end we'll probably use this to encrypt data with something like AES-256, which has a 256-bit key, so more entropy than that may not be useful.

rng/unix/getentropy.ml
```diff
@@ -0,0 +1,20 @@
+
+external getrandom_buf : bytes -> int -> int -> unit = "mc_getrandom" [@@noalloc]
```
edwintorok, line 2: I found this library: https://randombytes.cr.yp.to/security.html, which implements this ordered choice:

> librandombytes-kernel uses getrandom if it finds it, otherwise getentropy if it finds it, otherwise /dev/urandom. In the case of /dev/urandom, librandombytes-kernel waits at program startup for /dev/random to become readable; however, it skips this if the file /var/run/urandom-ready exists (or if /dev/random does not exist).

The wait on startup for /dev/random is probably not needed on modern Linuxes that wait for the RNG to be seeded (although I think very old kernels had that problem).

edwintorok: https://randombytes.cr.yp.to/librandombytes-20240318/kernel/getentropy_wrapper.c.html this has various workarounds for detecting the availability of getentropy by setting up a signal handler for SIGSYS (I assume some OS would raise a SIGSYS instead of returning ENOSYS).

rng/unix/urandom.ml
```diff
@@ -0,0 +1,27 @@
...
+  let ic = In_channel.open_bin "/dev/urandom"
+  and mutex = Mutex.create ()
+  in
+  (ic, mutex)
+
+let generate_into ~g:(ic, m) buf ~off len =
+  let finally () = Mutex.unlock m in
+  Mutex.lock m;
+  Fun.protect ~finally (fun () ->
+      match In_channel.really_input ic buf off len with
```
edwintorok, line 17: Looking at https://randombytes.cr.yp.to/librandombytes-20240318/kernel/devurandom_wrapper.c.html it does a `sleep(1)` whenever a read returns 0 on /dev/urandom.
I don't know whether that is just defensive programming (avoid a busy loop) or whether it was there to work around a bug on some OS (but I'd rather not see sleeps in crypto code).

rng/unix/mirage_crypto_rng_unix.ml
```diff
@@ -1,5 +1,21 @@
...
+let use_dev_urandom () =
+  let g = create (module Urandom) in
+  set_default_generator g
+
+let use_getentropy () =
+  let g = create (module Getentropy) in
+  set_default_generator g
+
+let use_default () =
+  try use_dev_urandom () with
```
edwintorok, line 16: https://randombytes.cr.yp.to/security.html suggests getentropy/getrandom is quite widely available, and avoids some of the bugs `/dev/urandom` might have.
OTOH testing in the linked issue has shown that `/dev/urandom` might be faster.

If performance is not a concern (during RSA key generation for example) then `getentropy` might be preferable?

rng/unix/mirage_crypto_rng_unix.mli
```diff
@@ -12,13 +12,24 @@ val initialize : ?g:'a -> 'a Mirage_crypto_rng.generator -> unit
 (** [getrandom size] returns a buffer of [size] filled with random bytes. *)
 val getrandom : int -> string
 
+(** A generator that opens /dev/urandom and reads from that file descriptor
+    data whenever random data is needed. The file descriptor is closed in
+    [at_exit]. *)
 module Urandom : Mirage_crypto_rng.Generator
 
+(** A generator using [getrandom(3)] on Linux, [getentropy(3)] on BSD and macOS,
+    and [BCryptGenRandom()] on Windows. *)
```
edwintorok, line 21: Should we recommend this as the default on modern OSes as a safe default (see `/dev/urandom` bugs on https://randombytes.cr.yp.to/security.html).
Although it is unclear whether using `getentropy` would've prevented various `/dev/urandom` bugs, like the one about initialization.

edwintorok:
> @edwintorok happy new year! It'd be great if you could have a look at this PR -- together with a release and announcement on discuss.ocaml.org, I think this would solve #249. What do you think?

Happy New Year!

I started looking at the PR, in general it looks good, but is a complex topic with many (historic) OS specific bugs.

Perhaps it might be useful to add a sanity  test that fills the buffer with a poison value, calls the RNG, and then checks that (most) bytes have been changed.
Although it is difficult to tell what a good threshold would be if we do this 256 times with each possible poison byte, then during at least one run we should detect that a cell has been changed.

I can try to submit such a test independently from this PR.

rng/unix/urandom.ml
edwintorok: Looking at the implementation: we use OCaml channels here, which reads IO_BUFFER_SIZE minimum https://github.com/ocaml/ocaml/blob/4.14/runtime/io.c#L103-L104
https://github.com/ocaml/ocaml/blob/e8d47983ea6b5556484caca3196b09b245f4e013/runtime/caml/io.h#L40

IO_BUFFER_SIZE is 65536 since 4.0 https://github.com/ocaml/ocaml/blob/trunk/Changes#L12812

So the value here should probably be 65536 to match what we have read anyway from `/dev/urandom`. Also confirmed with strace that reading just 1 byte with `In_channel.really_input_string` results in a syscall that reads 64KiB.

rng/unix/urandom.ml
```diff
@@ -0,0 +1,28 @@
+
+type g = In_channel.t * Mutex.t
+
+let block = 2048
```
edwintorok, line 4:
```suggestion
let block = 65536
```

rng/unix/getentropy.ml
```diff
@@ -0,0 +1,20 @@
...
+
+type g = unit
+
+let block = 256
```
edwintorok, line 6:
```suggestion
(* The maximum value for length is GETENTROPY_MAX : https://pubs.opengroup.org/onlinepubs/9799919799/functions/getentropy.html
    The minimum acceptable value for GETENTROPY_MAX is 256 https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/limits.h.html
  *)
let block = 256
```

rng/unix/urandom.ml
edwintorok: But anyway we can't do anything about that on the OCaml side, since short reads are handled by the OCaml runtime

rng/unix/mc_getrandom_stubs.c
```diff
@@ -72,7 +74,7 @@ void raw_getrandom(uint8_t *data, uint32_t len) {
 #error "Retrieving random data not supported on this platform"
 #endif
 
-CAMLprim value mc_getrandom (value buf, value len) {
-  raw_getrandom(Bytes_val(buf), Int_val(len));
+CAMLprim value mc_getrandom (value buf, value off, value len) {
+  raw_getrandom(_bp_uint8_off(buf, off), Int_val(len));
```
edwintorok, line 78: This is missing some bounds checks leading to segfaults:
```
utop # Mirage_crypto_rng_unix.use_getentropy ();;
- : unit = ()
─( 16:15:55 )─< command 13 >────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Mirage_crypto_rng.generate_into b ~off:(-10) 10;;
- : unit = ()
─( 16:16:09 )─< command 14 >────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Mirage_crypto_rng.generate_into b ~off:(-1000) 10;;
[2]    270243 segmentation fault  dune utop
```

Although there are *some* bounds checks in rng.ml they are insufficient: both `off` and `len` needs to be bounds checked to be between `[0, buffersize)`.

edwintorok: Similar segfault if length is exceeded:
```
─( 16:19:52 )─< command 10 >────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Mirage_crypto_rng.generate_into b 10000;;
Exception: Invalid_argument "buffer too short".
─( 16:19:53 )─< command 11 >────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Mirage_crypto_rng.generate_into b (-10000);;
[2]    271130 segmentation fault  dune utop
```

I think this needs some unit tests at the Rng level, and each Rng should then instantiate that test and check that bounds are properly checked for both parameters.

edwintorok, line 78: Also there are no upper bounds checks performed on `len`.
There is quite a lot of confusion with integer types, having used all of `uint32_t`, `size_t`, `ssize_t` and `int`. I can't quite wrap my head around what the result of integer promotion will be from that, so I tested it instead.

If you supply too big of an argument to `len` on a 64-bit build then you'll get back a buffer with a bunch of zeroes, which is clearly not random...:

utop # Mirage_crypto_rng.generate (1 lsl 33);;
- : string =
"\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000"... (* string length 8589934592; truncated *)

I think this should probably be bounds checked after calling `Int_val` to check that:
* the result is not negative (e.g. an overflow on 32-bit builds because the argument is uint32_t but the full range of that is not representable on 32-bit builds)
* the result doesn't exceed the maximum value representable in an uint32_t

Although it is unlikely that someone would call the function with such out of range values, crypto code should be robust and raise exceptions rather than return uninitialized data.

dinosaure: Can be done on the OCaml side I think (otherwise, if the C stub is able to raise an exception, we loose the `[@noalloc]`).

rng/unix/mirage_crypto_rng_unix.mli
edwintorok: Although seeing just how error prone writing a binding to `getentropy` is (it is quite easy to forget adding bounds checks, which can then result in segfaults or use of uninitialized data) I'd rather stick with `geturandom` as the default.

AFAICT kernel bugs affect both `/dev/urandom` and `getentropy`/`getrandom` the same way, the only difference is that the latter is available in restricted environments like chroots.

edwintorok:
BTW I started writing some RNG tests last year: https://github.com/perf101/rage/tree/private/edvint/stats6/test/rngtest
It was for a different purpose, I was looking for good quality uniform randomness, but that is a minimum requirement for cryptographic RNGs (of course non-cryptographic RNGs, like the one used by OCaml in 5.x are much faster, and actually have very good randomness properties, so that is what I picked for that particular project).

Although running the full suite takes several hours (and there are multiple testsuites where experts disagree which one is better, so I used them all: `dieharder`, `TestU01`, and `practrand`. And due to way tests work a "true" RNG is expected to fail these tests some of the time (in fact exactly as many times as its `p` value says, which can't be zero because then it won't reject bad RNGs either), and you can have a meta test that then looks at the distribution of these failures, but that needs to have a p-value to fail, and so on, so using this in a CI might be tricky, but a suitable p-value and a suitable number of repetitions for the "meta" test might result in something usable in a CI.
It should be possible to have a cut down version of those tests that finish in a couple of minutes that could be used as a sanity test in the CI, I'll see whether I can put together a PR for that (I currently lack the 'meta' testing for some of the tests).

rng/unix/mc_getrandom_stubs.c
hannesm: Thanks for this test, in 520732b I modified the C code to use size_t instead of uint32_t -- does this make the issue disappear on your machine?

On my laptop, I get OOM-killed when trying to use getrandom with len = 1 lsl 33.

hannesm: I check in fafabab that off and n are positive. This is sufficient for the API you use -- there's the niche usage of the direct generators (e.g. `Mirage_crypto_rng.Fortuna.generate_into`) that doesn't do appropriate bounds checking. I think I'll remove that from the public API.

rng/unix/mirage_crypto_rng_unix.ml
hannesm: I'm on the fence with that. The current code uses dev_urandom due to the performance. But I think using getentropy/getrandom as default is also a valid choice.

In terms of performance, indeed it shouldn't matter too much. If your application has the RNG as a bottleneck, you're able to manually select the thing you like within that application.

rng/unix/mc_getrandom_stubs.c
hannesm: FWIW, I pushed 5919c92 which emits alerts.

rng/unix/mirage_crypto_rng_unix.ml
hannesm: Done in 74e4e67

rng/unix/mirage_crypto_rng_unix.mli
hannesm: I decided we default to getrandom/getentropy/... in 74e4e67

hannesm:
CI failures:

OCaml-CI:
- openbsd: some core_unix/base_unix issues, won't fix
- debian 5.3 beta 1: async_rpc_kernel compilation issue, won't fix
- windows 5.3:
``` 
File "tests/dune", line 72, characters 26-54:
72 |  (names     test_miou_rng test_miou_entropy_collection)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(cd _build/default/tests && .\test_miou_entropy_collection.exe)
Fatal error: exception Unix.Unix_error(Unix.ENOTSOCK, "caml_unix_set_nonblock", "")
File "tests/dune", line 72, characters 12-25:
72 |  (names     test_miou_rng test_miou_entropy_collection)
^^^^^^^^^^^^^
(cd _build/default/tests && .\test_miou_rng.exe)
Fatal error: exception Unix.Unix_error(Unix.ENOTSOCK, "caml_unix_set_nonblock", "")
```

So, some issue with miou and windows.

The github windows action: issues with gmp.

TL;DR: CI errors are unrelated to the proposed change.

The uint32_t -> size_t change awaits feedback from Edwin. I think otherwise the PR is fine to merge.

rng/unix/mc_getrandom_stubs.c
edwintorok: Ah Int_val will truncate, I think you want Long_val.

edwintorok, line 78:
```suggestion
  raw_getrandom(_bp_uint8_off(buf, off), Long_val(len));
```

edwintorok: With that change I now get a proper output:
```
utop #  Mirage_crypto_rng.generate (1 lsl 34);;
- : string =
"#\ba¢¦©°¢\031¤·ür¥½=7(ca`JµÙÇÃÉiÂ3ï»áj&\"Ó\014%÷\007J-)ó|oì\016Û&z9O\028+@»\007M@\b\002òõÆþ/¼\019æ\023»+%O0³Ò\r©ê°\022Hç*#\rê$Ï;ÊlúÙ2ÿ;A`\028'ú6ã{Âölm´\026\\Ƞë\001u©î.uS7÷\031:\017CTú³?¡:­=ácI]¸t£h¬^Qëq\023PöæQ\026û9\007§68r®)LZo⨻&ã\127ýÙ\031fCßáVȒqó\025roÂýv!Ã5öÉ\"ñµ ޅJ\tý\020sUoaßѸ\000;Øn²Ë\031=­M\127n\127\028\n:=`­md}÷nMäåZ8\nۺ¯kÑ\025hhû\015åêם&º3"... (* string length 17179869184; truncated *)
─(
```

edwintorok: Thanks, the bounds checks look good.

edwintorok:
> The uint32_t -> size_t change awaits feedback from Edwin

There is one missing piece there: `Int_val` also needs to be changed to `Long_val`, see the latest suggested change on github.

If you make that change I think this PR is ready then, thanks a lot for the improvements.

rng/unix/mc_getrandom_stubs.c
hannesm: I applied your suggestion.Thanks.

edwintorok (approved):

hannesm:
Ok, I like the diff. One remaining question is whether we should immediately deprecate the Mirage_crypto_rng_unix.initialize, Mirage_crypto_rng_lwt.initialize, Mirage_crypto_rng_async.initialize, Mirage_crypto_rng_miou.initialize (also Mirage_crypto_rng_miou_unix.iniitialize) and point to `Mirage_crypto_rng_unix.use_default ()`?

Maybe @dinosaure and @reynir have ideas about that. It also doesn't block merging this PR, only blocks a subsequent release. It'd ease the opam package cone as well (since in the end we could remove these packages, and only retain mirage-crypto-rng and mirage-crypto-rng-mirage).

Issue #251, opened 2025-01-07
rdseed/rdrand reject true 0 result


edwintorok:
While testing the various RNGs/Entropy sources I got a failure quite often with `Mirage_crypto_rng.Entropy: 0 is a bad CPU RNG value`.

Reading the manual this is wrong, only the CF can be used to determine success or failure, NOT the return value: 
https://www.intel.com/content/www/us/en/developer/articles/guide/intel-digital-random-number-generator-drng-software-implementation-guide.html

> Note that a destination register value of zero should not be used as an indicator of random value availability. The CF is the sole indicator of the success or failure of the RDRAND instruction.

Which makes sense: 0 is a valid return value, one that a RNG is expected to generate at least some of the time, if it wouldn't then the generator would be biased and would cause various RNG tests to fail.

edwintorok:
> * with RETRIES being 10)

What if `rdseed` still fails after more than 10 retries (quite easy to trigger that, unlike for `rdrand` where triggering many failures in a row is not that easy).
It'll return 0, which is indistiguishable from a "successful" 0.

If we don't want to allocate/raise exceptions then perhaps passing in an OCaml bool ref that can be modified to signal an error could work?

edwintorok:
Another idea: the stub could have return type `bool`, and get a bytes buffer as argument that can be filled in with the seed.
That way we retain *all* the bits, including the one we are currently losing for the GC integer tag.

hannesm:
word! that's what we do!

hannesm:
Fixed in #255

PR #255, opened 2025-01-08
Handle CPU_RNG failures in OCaml:

hannesm:
- Previously, a return value of "0" was treated specially, but rdrand/rdseed may return 0!
- If there's a failure in rdrand/rdseed during cpu_rng_bootstrap, raise an exception
- If there's a failure during the periodic reseeding (timer initiated feeding of all pools), ignore it (for now!?)

//cc @edwintorok addresses #251 -- though I'm not yet happy with the `cpu_rng` failure case // `feed_pools` error case. I'm uncertain what to do:
- do a while true, potentially leading to a busy loop if rdrand/rdseed keeps on giving no data
- raise an exception and exit - which may be inconvenient for the application developer
- call rdseed/rdrand (the other instruction) as an attempt to delay the error condition

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -205,8 +205,6 @@ enum cpu_rng_t {
 
 static int __cpu_rng = RNG_NONE;
 
-#define RETRIES 10
-
 static void detect (void) {
 #ifdef __mc_ENTROPY__
   random_t r = 0;
@@ -214,7 +212,7 @@ static void detect (void) {
   if (mc_detected_cpu_features.rdrand)
     /* AMD Ryzen 3000 bug where RDRAND always returns -1
        https://arstechnica.com/gadgets/2019/10/how-a-months-old-amd-microcode-bug-destroyed-my-weekend/ */
-    for (int i = 0; i < RETRIES; i++)
+    for (int i = 0; i < 10; i++)
       if (_rdrand_step(&r) == 1 && r != (random_t) (-1)) {
         __cpu_rng = RNG_RDRAND;
         break;
@@ -223,7 +221,7 @@ static void detect (void) {
   if (mc_detected_cpu_features.rdseed)
     /* RDSEED could return -1, thus we test it here
        https://www.reddit.com/r/Amd/comments/cmza34/agesa_1003_abb_fixes_rdrandrdseed/ */
-    for (int i = 0; i < RETRIES; i++)
+    for (int i = 0; i < 100; i++)
       if (_rdseed_step(&r) == 1 && r != (random_t) (-1)) {
         __cpu_rng |= RNG_RDSEED;
         break;
@@ -235,7 +233,7 @@ CAMLprim value mc_cpu_rdseed (value buf, value off) {
 #ifdef __mc_ENTROPY__
   random_t r = 0;
   int ok = 0;
-  int i = RETRIES;
+  int i = 100;
   do { ok = _rdseed_step (&r); _mm_pause (); } while ( !(ok | !--i) );
   fill_bytes(buf, off, &r);
   return Val_bool (ok);
@@ -249,7 +247,7 @@ CAMLprim value mc_cpu_rdrand (value buf, value off) {
 #ifdef __mc_ENTROPY__
   random_t r = 0;
   int ok = 0;
-  int i = RETRIES;
+  int i = 10;
   do { ok = _rdrand_step (&r); } while ( !(ok | !--i) );
   fill_bytes(buf, off, &r);
   return Val_bool (ok);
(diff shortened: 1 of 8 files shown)


src/native/entropy_cpu_stubs.c
```diff
@@ -229,29 +245,31 @@ static void detect (void) {
...
 
-CAMLprim value mc_cpu_rdseed (value __unused(unit)) {
+CAMLprim value mc_cpu_rdseed (value buf, value off) {
 #ifdef __mc_ENTROPY__
   random_t r = 0;
   int ok = 0;
   int i = RETRIES;
   do { ok = _rdseed_step (&r); _mm_pause (); } while ( !(ok | !--i) );
-  return Val_long(r);
+  fill_bytes(buf, off, r);
```
edwintorok, line 254: Why not `memcpy` from `&r`? IIRC GCC usually optimizes the call away for known static sizes, although this can be confirmed by looking at the generated assembly.

hannesm: no reason, just stupidity. fixed in 21945e6

hannesm:
Thanks for your insights. So, for `cpu_rng_bootstrap`, I think failing hard is fine (this is done at startup anyways). This is as well the thing that prefers to call rdseed.

Now, for periodic feeding of the rng, we use the `cpu_rng` task - which prefers to use rdrand! Here the failure is currently ignored (and indeed, lwt/async/eio/miou_unix call this periodically -- by default every second.

My questions are now:
- if the cpu only has rdrand, should our cpu_rng_bootstrap do more work (call rdrand 1024 times and compress the random, as done in the reference you linked to)?
  - related: should if rdseed fails in cpu_rng_bootstrap we go the rdrand route (with multiple calls) instead of raising an exception?
- if the cpu only has rdseed, should our periodic task handle that more graceful?
- should we increase RETRIES for rdseed to 100?

edwintorok:
> this is done at startup anyways)

Startup of a mirage unikernel perhaps, but not startup of the host. There could be any number of other VMs or processes running on the host CPU already, and I think depleting entropy with RDSEED depletes it globally for the whole CPU, not just your core, but I'll have to do some tests to confirm that (i.e. if I busy loop calling rdseed or rdrand on 31 cores, will my 32th core get a chance of a successful rdseed ever?)



> * and compress the random,

I think that is already taken care by Fortuna. If you give Fortuna all the rdrand values you have it'll figure out what to do with them (IIRC Fortuna doesn't require you to estimate how much entropy you've got in each sample).



> * f the cpu only has rdrand, should our cpu_rng_bootstrap do more work (call rdrand 1024 times 
>   
>   * related: should if rdseed fails in cpu_rng_bootstrap we go the rdrand route (with multiple calls) instead of raising an exception?
> * if the cpu only has rdseed, should our periodic task handle that more graceful?
> * should we increase RETRIES for rdseed to 100?

I'd say probably, but I'll need to reread the manual and do some experiments (also I'll need to see whether the AMD SDM has any documentation on this that has other suggestions). 

When I looked at this yesterday 100 wasn't enough for RDSEED. Reading between the lines it seems like it'd be possible for some cores to DoS the ability to run RDSEED on other cores ~forever, but I'll need to confirm whether that is true, and whether in that case falling back to RDRAND would help or not.

But regardless of what the manual says we can't trust hardware (and future generations of hardware may behave differently or have different bugs anyway), so we must have some way to handle failure.

I'd say that a failure during reseeding is not critical: if it was safe enough for the RNG to work and generate numbers before, then why wouldn't it be safe to continue as is. And if the answer is that it isn't safe to continue without reseeding then the RNG wasn't seeded well enough in the first place and shouldn't have given out any random numbers until properly seeded.

I'd suggest to use an Atomic.incr to count the number of failures, and someone interested could monitor that counter, set up some alerting from the application, or crash the application if that is what they wish, but the library should leave that choice up to the application.

hannesm:
Thanks again, edwin. With 276ffe2 the rdrand and rdseed calls and failures are exposed now.

Remaining: rdrand when rdseed is unavailable during bootstrap; also resort to rdrand if rdseed is not returning good values?

And how many RETRIES for rdseed.

hannesm:
Thank you very much for your time, comments and testing!

hannesm:
With a55afa4 if only rdrand is available or rdseed fails during bootstrap, we call it 512 times.

An error is logged on failure.

The code, as before (in bootstrap) - resorts to using whirlwind.

hannesm:
I adjusted the retries -- rdseed now 100, rdrand still at 10.

From my side, this is ready for review & merge. It may be sensible to cut a release thereafter (with #254 merged as well).

Any opinion @edwintorok?

dinosaure (approved):
Seems ok, the errors on the CI seems unrelated.

reynir (approved):

PR #259, opened 2025-03-04
mirage-crypto-ec: implementation of SECP256K1

ansiwen:
This change implements the SECP256K1 curve (also known as the Bitcoin curve).
 - field primitives are generated by the fiat-crypto project[1]
 - point primitives are generated by the ECCKiila project[2]
 - Ocaml point operations are taken from NIST implementation, adapted to ECCKiila point primitives and optimized for a=0.
 - testvectors for ECDH and ECDSA verification from wycheproof[3]

[1] https://github.com/mit-plv/fiat-crypto
[2] https://gitlab.com/nisec/ecckiila
[3] https://github.com/C2SP/wycheproof


ansiwen:
CI failures seem to be unrelated.

ansiwen:
FWIW: I will also implement Brainpool curves. Please let me know, if you are interested in these as well.

hannesm:
Thank you for your PR. Now, I have some questions:
- what license is ECCkiila, is it fine to embed the code?
- could you follow the other curve implementations and have the code generated by fiat added to the ec/native/GNUmakefile, so we can at any point when fiat updates generate these C files?
- where are the unit tests taken from? if from a RFC or elsewhere, could you add a remark in the file?
- the changes to mirage_crypto_ec.ml are substantial, could you please minimize the diff (i.e. changing the constructor of field_element, renaming b_uts to out_fe_to_fe, rename Foreign to Foreign_field etc. -- this can be done in a separate PR, but please keep these things separate)

> As an alternative I can also migrate the NIST curves to ECCKiila code, which is about 10% faster and would make the code simpler and more unified.

Is this true across CPU architectures, or only on your Apple silicon? As you may have noticed, the other EC code is taken from projects with verification in mind, how much review and formal verification did ECCkiila get?

> FWIW: I will also implement Brainpool curves. Please let me know, if you are interested in these as well.

Sure, in the end it all depends on code clarity and binary size (we dropped P224 for the binary size reason).

ansiwen:
> * what license is ECCkiila, is it fine to embed the code?

The license (MIT) is included in the generated file: https://github.com/mirage/mirage-crypto/blob/1a4e3f20ac8e4f791ce0213c62740ddf8fc37a03/ec/native/secp256k1.h#L1-L24

> * could you follow the other curve implementations and have the code generated by fiat added to the ec/native/GNUmakefile, so we can at any point when fiat updates generate these C files?

The fiat code is automatically embedded in the file generated by ECCKiila. So, you want me to create the fiat files separately and remove them from the generated ECCKiila file? I sure can do that, just want to make sure we are on one page.

> * where are the unit tests taken from? if from a RFC or elsewhere, could you add a remark in the file?

Unfortunately there seem to be any official test vectors for ECDSA and secp256k1, so I frankensteined them from some single test vectors I found at different places. (One of them is from Bouncy Castle IIRC)

> * the changes to mirage_crypto_ec.ml are substantial, could you please minimize the diff (i.e. changing the constructor of field_element, renaming b_uts to out_fe_to_fe, rename Foreign to Foreign_field etc. -- this can be done in a separate PR, but please keep these things separate)

Ok, but some code I had to change in order to make it reusable. Or would you prefer code duplication and untouched old code?

> > As an alternative I can also migrate the NIST curves to ECCKiila code, which is about 10% faster and would make the code simpler and more unified.
> 
> Is this true across CPU architectures, or only on your Apple silicon?

I just tested it on x86_64, there it is also around 10%.

> As you may have noticed, the other EC code is taken from projects with verification in mind, how much review and formal verification did ECCkiila get?

You mean other projects than fiat-crypto? They don't have any formal verification beyond fiat. From their [paper](https://arxiv.org/pdf/2007.11481):

> It is important to note [Fiat] is the formal verification boundary for ECCKiila—all other code on top of Fiat, while computer generated through templating and automatic formula generation for ECC arithmetic, has no formal verification guarantees.

If the current EC arithmetic in mirage-crypto has stronger guarantees, that is of course an argument to keep the implementation, despite the slight performance difference.

Let me know what you prefer, and I will proceed accordingly.

ansiwen:
I reduced the diff as much as possible without creating too much duplicate code, and split it up in two commits. The first is a pure restructuring without functional difference. The second is the addition of the new curve. Hope that helps to digest it.

ansiwen:
> > As you may have noticed, the other EC code is taken from projects with verification in mind, how much review and formal verification did ECCkiila get?
> 
> You mean other projects than fiat-crypto? They don't have any formal verification beyond fiat. From their [paper](https://arxiv.org/pdf/2007.11481):
> 
> > It is important to note [Fiat] is the formal verification boundary for ECCKiila—all other code on top of Fiat, while computer generated through templating and automatic formula generation for ECC arithmetic, has no formal verification guarantees.
> 
> If the current EC arithmetic in mirage-crypto has stronger guarantees, that is of course an argument to keep the implementation, despite the slight performance difference.

I looked a bit more into the point add and double formulas ECCKiila is using for code generation. They are taken from [this paper](https://eprint.iacr.org/2015/1060.pdf) (which is also referenced in the comments of the generated code) that provides at least Magma code which provides "verification of the complete addition and exception-free doubling formulas". So there seems to be indeed some verification of the algorithm itself, and I guess if the tooling has no bugs, this is also true for the generated C code? Unfortunately I can't really tell how that compares to the guarantees of the implementations in `point_operations.h`.

ansiwen:
FWIW: I did more comparisons of the ECCKiila implementation with the current implementation on x86_64 and arm64:

x86_64 (Intel Core i7, 2.6GHz)

  | current | ECCKiila | % | ECCKiila u-sol | %
-- | --: | --: | --: | --: | --:
P256 DSA sign | 5504 | 6023 | 109 % |   |  
P256 DSA verify | 1004 | 1645 | 164 % |   |  
P256 DH share | 1261 | 2344 | 186 % |   |  
P256 DH secret | 7964 | 9156 | 115 % |   |  
P521 DSA sign | 773 | 796 | 103 % | 2519 | 326 %
P521 DSA verify | 108 | 191 | 177 % | 1010 | 935 %
P521 DH share | 137 | 252 | 184 % | 1751 | 1278 %
P521 DH secret | 883 | 918 | 104 % | 4302 | 487 %


arm64 (MacBook M3)

  | current | ECCKiila | % | ECCKiila u-sol | %
-- | --: | --: | --: | --: | --:
P256 DSA sign | 15718 | 17330 | 110 % |   |  
P256 DSA verify | 2990 | 5335 | 178 % |   |  
P256 DH share | 3834 | 7534 | 197 % |   |  
P256 DH secret | 24241 | 28464 | 117 % |   |  
P521 DSA sign | 2901 | 2926 | 101 % | 4957 | 171 %
P521 DSA verify | 429 | 754 | 176 % | 1529 | 356 %
P521 DH share | 542 | 1044 | 193 % | 2243 | 414 %
P521 DH secret | 2822 | 2835 | 100 % | 4659 | 165 %

The "u-sol" column is a ECCKiila with unsaturated-solinas fiat code for field arithmetic. (scalar arithmetic is still montgomery, couldn't create unsaturated-solinas for the order of P-521, probably wouldn't be that efficient anyway.)

As you can see, there are some vast improvements for P-521, but the unsaturated-solinas could probably also be used independently of ECCKiila, I guess.

ansiwen:
I reorganized the code and created a bunch of new packages, which allow to selectively include only the curves that are required. See my branch here: https://github.com/Nitrokey/nethsm-mirage-crypto/tree/nethsm-crypto

Unfortunately, after all that work I realized, that we don't use `Mirage_crypto_ec` directly, but through the `X509` package, which seems to be highly entangled with Mirage_crypto. To make `X509` to support other curves from third party libraries as a kind of plugin looks like lot of work, and I even don't know if this would be something that upstream would be interested in. WDYT? Is that something the upstream projects would like to go?

Otherwise at the moment it looks like a lot more feasible to just extend `Mirage_crypto` and `X509` to me, as a kind of "extended fork", which would be quite straight forward.

PR #260, opened 2025-03-10
mirage-crypto-ec: add Brainpool curves with 254/384/512 bits

ansiwen:
This change adds the Brainpool curves, namely brainpoolP254r1,
brainpoolP384r1, brainpoolP512r1.  The implementation internally uses the
twisted versions of the curves (...t1) in order to use the same a=-3 EC
arithmetic as the NIST curves.  Point coordinates in regular form are
transformed to twisted form before calculations, and vice versa afterwards.

Based on: #259, merge that first


ansiwen:
Setting to "ready for review" although some CI tests are failing. Not sure why. Any clue how [this](https://github.com/mirage/mirage-crypto/actions/runs/13771523477/job/38511047341?pr=260#step:6:121) can happen, but only for 4.14.2 on ubuntu and windows? Is there maybe a cache issue (because the file was indeed missing before)?

ansiwen:
> Setting to "ready for review" although some CI tests are failing. Not sure why. Any clue how [this](https://github.com/mirage/mirage-crypto/actions/runs/13771523477/job/38511047341?pr=260#step:6:121) can happen, but only for 4.14.2 on ubuntu and windows? Is there maybe a cache issue (because the file was indeed missing before)?

Answering myself: Mac is case-insensitive, and the files in `test_ec_wycheproof.ml` had lower-case `p`.

PR #266, opened 2025-05-26
[WIP] Add `mirage-crypto-rng-miou-solo5`

dinosaure:
To be able to keep tracking this new package, I make a PR which shows what is actually required to be able to integrate `mirage-crypto-rng` into [`miou-solo5`](https://github.com/robur-coop/miou-solo5). We currently have two issues:
- [ ] we must inhibit the compilation without acceleration. On my machine, it's seems that `ocaml-solo5` does not provide everything needed for some GCC instructions. Not sure if it's related to the last version of GCC.
- [ ] try to add a hook which performs `Entropy.timer_accumulator` degrades a lot performance on a basic unikernel made with `miou-solo5`. We should clarify the strategy and what is needed to have a good RNG on `miou-solo5`.


hannesm:
Acceleration: this is strange -- since the "normal" freestanding build does use them and seems to work. I'm not sure why this fails on your machine / in your scenario. Would you mind to report the error that you discover in this case?

Entropy harvesting is hard -- the strategy used (for e.g. MirageOS and Lwt) is to use some bits of the clock to be executed at each Lwt loop enter -- this was performance-evaluated to take only a couple of nanoseconds back then. In addition, if available, CPU instructions (rdrand) are used to feed entropy). Now, I'm surprised that on miou-solo5 it degrades the performance a lot. :/ From my point of view, only relying on rdrand is not a good idea - it is unclear how good these instructions are (plus on some CPUs like AMD Ryzen, there were issues with these instructions). Do you have an idea why the performance is degraded that much?

src/native/entropy_cpu_stubs.c
```diff
@@ -9,15 +9,23 @@
...
 #if defined (__i386__) || defined (__x86_64__)
-#include <x86intrin.h>
 
-#if defined (__x86_64__)
+#if defined (__ocaml_solo5__)
+inline uint64_t __rdtsc(void) {
+  unsigned l, h;
+  __asm__ __volatile__("rdtsc" : "=a"(l), "=d"(h));
+  return ((uint64_t)h<<32) | l;
+}
```
hannesm, line 18: this change seems to make the mirage test on FreeBSD fail (due to conflicting types of __rdtsc.

dinosaure:
PS: note that `rdrand` (provided by `mirage-crypto-rng`) is executed every 1 second.

hannesm:
The rough idea for Lwt is to harvest some entropy from every external event (i.e. a network packet arrived etc.) -- this means whenever the scheduler is executed (see https://ocsigen.org/lwt/latest/api/Lwt_main.Enter_iter_hooks).

Now, with miou its scheduler is executed more often, as I understand. It would be great to have the possibility to hook whenever something is added to (or something is removed from) the set of tasks to be scheduled. A periodic task "every 1s, get the time and feed it to the entropy pool" is too deterministic, and would defeat the purpose.

dinosaure:
I just tested my last commit and it seems that the hook does not degrade too much performances (with `iperf3`). So one issue is solved :tada:. I will dig in latter about acceleration.

rng/miou/mirage_crypto_rng_mkernel.ml
```diff
@@ -0,0 +1,60 @@
...
+
+let running = Atomic.make false
+
+let default_generator_already_set =
+  "Mirage_crypto_rng.default_generator has already \
+   been set (but not via Mirage_crypto_rng_miou_solo5). Please check \
+   that this is intentional"
+
+let miou_generator_already_launched =
+  "Mirage_crypto_rng_miou_solo5.initialize has already been launched \
```
hannesm:
```suggestion
  "Mirage_crypto_rng_mkernel.initialize has already been launched \
```

PR #267, opened 2025-06-18
Add a missing dependency in the configuration detection

shym:
This PR fixes an issue due to a missing dependency on the configuration for the configuration detection.

The configuration detection executable uses `dune-configurator` and so it uses the configuration passed in (currently) `.dune/configurator.v2` in particular to know which C compiler it should invoke. This adds an explicit dependency to this file (with a glob in case the format changes) to make sure a change of compiler triggers a redetection.

The issue arose when building a MirageOS unikernel with a Unikraft backend. The OCaml/Unikraft packaging makes it easy to switch between arm64 and x86_64 architecture, where only the `%{lib}%/findlib.conf.d/unikraft.conf` file is updated to point to one or the other toolchain. With such a change of target, `dune` was missing the dependency information to rerun `cfg.exe` with the proper C compiler.

This PR also takes the opportunity to centralise the detection in one directory.

The bug investigation is joint work with @Firobe.

### DIFF
--- a/config/dune
+++ b/config/dune
@@ -1,3 +1,8 @@
 (executables
  (names cfg)
  (libraries dune-configurator))
+
+(rule
+ (targets cflags.sexp cflags_optimized.sexp cflags_warn.sexp)
+ (action
+  (run ./cfg.exe)))
--- a/config/dune
+++ b/config/dune
@@ -4,5 +4,7 @@
 
 (rule
  (targets cflags.sexp cflags_optimized.sexp cflags_warn.sexp)
+ (deps
+  (glob_files %{workspace_root}/.dune/configurator*))
  (action
   (run ./cfg.exe)))
--- a/ec/dune
+++ b/ec/dune
@@ -9,15 +9,10 @@
   (include_dirs ../src/native)
   (flags
    (:standard -DNDEBUG)
-   (:include cflags_optimized.sexp))))
+   (:include ../config/cflags_optimized.sexp))))
 
 (env
  (dev
-  (c_flags (:include cflags_warn.sexp))))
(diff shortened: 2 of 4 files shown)


hannesm:
Thanks, though I'm curious -- I copied this detection logic from elsewhere (I don't remember where), and I've never seen this

```
 (deps
  (glob_files %{workspace_root}/.dune/configurator*))
```

Is this a new thing? Is this documented somewhere? I couldn't find anything e.g. in https://dune.readthedocs.io/en/stable/dune-libs.html#configurator

Also, I'm not sure about workspace_root (never used that), neither what kind this ".dune/configurator*" is? An executable? A directory?

I find this rather mysterious, and would like to understand why this is needed and which other projects need these magic lines?

hannesm:
I digged into the mirage issue tracker, and is this one relevant https://github.com/mirage/mirage/issues/1195#issuecomment-726017019 -- where some changes were needed to some of the packages that used the dune-configurator to play nicely with opam-monorepo / the MirageOS 4.0 story of building unikernels using cross-compilation?

hannesm:
From the linked dune issue, it sounds like there's a different solution the dune developers have in mind, but also they haven't done any action to move this forward. What is the state of this PR and dune changes?

shym:
On the dune side, I expect that this is waiting on _someone_ having some spare time to implement that. I hope to get round to do this, but I’ve no idea when I’ll find some time to look into that.
So I see 2 possibilities:
- merge the PR as is, knowing it’s a hack that should be rewritten whenever dune provides cleaner support for this,
- keep this PR open, or turn it into an issue, to keep some track of the underlying problem.

PR #268, opened 2025-08-07
mirage-crypto-ec: fix ECDSA public key compression

ansiwen:
The (de-)compression code for ECDSA public keys doesn't use the least significant byte for extracting the parity bit, but the second-least instead, which results in the wrong `0x02`/`0x03` prefix with 50% chance. (Probably the result of some former refactoring.) Because de- and compression code is consistently wrong, the test didn't fail.

This change fixes the compression code, adds an independent prefix check in the compression test and repeat the test 20 times to make sure it hits.

### DIFF
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -317,7 +317,7 @@ module Make_point (P : Parameters) (F : Foreign) : Point = struct
     if compress then
       let out = Bytes.create (P.byte_length + 1) in
       let ident =
-        2 + (String.get_uint8 buf ((P.byte_length * 2) - 1)) land 1
+        2 + (String.get_uint8 buf (P.byte_length * 2)) land 1
       in
       Bytes.unsafe_blit_string buf 1 out 1 P.byte_length;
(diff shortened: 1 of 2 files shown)


hannesm:
Thanks for your PR. I don't have much insight into the EC point compression, and would be happy if you could provide test vectors from the specification, to avoid any future issues.

ansiwen:
> > prefix it with the byte 02 if the Y coordinate is even, or 03 if the Y coordinate is odd
> 
> > in a single test vector you would still have a 50% chance
> 
> thanks for being pedantic. maybe two test vectors, one with an odd Y coordinate and one with an even Y coordinate, would solve the issue?

No, it would still have a 25% chance of being a false positive, because the bug is that it's not extracting the correct bit, and you can't assume which one it is in a general test. I can easily construct two such test vectors which the old code would pass. I have no idea how to cover all possibilities systematically.

ansiwen:
We could test with a Y that is ...00000001 and ...11111110, but these values are not on the curve, so we would need to skip the coordinate validation for the test then. Not sure if it's worth the hassle.

PR #273, opened 2026-02-13
Disable test_entropy for riscv 64

reynir:
This addresses #272.

There are unrelated fatal unused values warnings in `ec/gen_tables/gen_tables.ml` that make it fail. But `test_entropy` is not run on riscv 64 bit.

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -35,7 +35,8 @@
  (modules test_entropy)
  (package mirage-crypto-rng)
  (libraries mirage-crypto-rng ohex)
- (enabled_if (and (<> %{architecture} "arm64") (<> %{architecture} "riscv64"))))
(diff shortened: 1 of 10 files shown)


tests/dune
```diff
@@ -35,7 +35,8 @@
  (modules test_entropy)
  (package mirage-crypto-rng)
  (libraries mirage-crypto-rng ohex)
- (enabled_if (and (<> %{architecture} "arm64") (<> %{architecture} "riscv64"))))
+ (enabled_if (and (<> %{architecture} "arm64")
+                  (not (and (= %{architecture} "riscv") %{arch_sixtyfour})))))
```
hannesm: maybe it is sufficient to say `<> %{architecture} "riscv"` -- so we disable this on riscv entirely?

PR #274, opened 2026-02-17
Use "architecture" "riscv" to fix the issue to not run on riscv64 #272

hannesm:
as suggested in #273 by @reynir


reynir (approved):
Thanks, I think this is sensible and avoids upgrading the dune version.

PR #278, opened 2026-03-10
add primitives for Thomas' spake2

hannesm:
alternative to #277


samoht:
Thanks, this looks much cleaner and simpler and solves indeed my use-case! Many thanks

hannesm:
I revised the commit message, will merge and cut a release.

PR #279, opened 2026-05-17
Skip entropy timer test on x86_64 macOS

iamanaws:
`test_entropy` can fail intermittently on x86_64 macOS because it expects consecutive cycle-counter samples to differ.
This updates the Dune `enabled_if` condition to skip that test on macOS `amd64`, matching the existing platform-specific skips for arm64 and riscv.

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -35,7 +35,12 @@
  (modules test_entropy)
  (package mirage-crypto-rng)
  (libraries mirage-crypto-rng ohex)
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks for your PR. Since the entropy collection is only used for MirageOS unikernels, I guess it is fine. On the other hand, I basically gave up on such a fitness function. Maybe there should be such a test at startup if entropy harvesting is used...

PR #281, opened 2026-06-25
chacha20: reject negative offsets in the AEAD bounds check

### LINKED ISSUE #280: A negative offset bypasses the AEAD bounds check
`Mirage_crypto.Chacha20.authenticate_encrypt_into` accepts a negative `src_off` that slips past its bounds check and reaches the noalloc C stub as a negative pointer offset. A large negative offset hits unmapped memory and the process dies with SIGSEGV; a small one xors memory next to the buffer into the ciphertext, leaking process memory. A bounds-checked API must reject a negative offset rather than forward it to native code. The defect is reachable through the public API.

To reproduce it:

```ocaml
let () =
  let key = Mirage_crypto.Chacha20.of_secret (String.make 32 'k') in
  let nonce = String.make 12 'n' in
  let src = String.make 16 'A' in
  let dst = Bytes.make 4096 '\000' in
  Mirage_crypto.Chacha20.authenticate_encrypt_into ~key ~nonce src  ~src_off:(-(1 lsl 38)) dst ~dst_off:0 ~tag_off:16 16;
  print_string "RETURNED NORMALLY (unexpected)\n"
```

Returns:

```
exit=139
```

The fix is to add a non-negativity guard at the top of `authenticate_encrypt_into` (and the matching `authenticate_decrypt_into`).

samoht:
Fix  #280.

### DIFF
--- a/src/chacha20.ml
+++ b/src/chacha20.ml
@@ -106,6 +106,8 @@ let unsafe_authenticate_encrypt_into ~key ~nonce ?(adata = "") src ~src_off dst 
   mac_into ~key:poly1305_key ~adata (Bytes.unsafe_to_string dst) ~src_off:dst_off len dst ~dst_off:tag_off
 
 let authenticate_encrypt_into ~key ~nonce ?adata src ~src_off dst ~dst_off ~tag_off len =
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks.

PR #282, opened 2026-06-29
Add Mirage_crypto_rng.entropy_test for testing entropy harvesting

hannesm:
In an application that relies on entropy harvesting, better test at startup whether the code works nicely. In case it doesn't, raise an exception. This way we can ensure to produce nice random numbers, and not fail silently.

//cc @dinosaure who may have an opinion about that. From my point of view, better fail hard and early. Plus there's some architectures / environments where some of the code actually needs to be implemented.

The other reason behind this is that the test_entropy test is disabled on more and more platforms since there is lots of stuff around that is broken or not implemented. The overall strategy to verify at compile time (when the tests are run) whether the build host is able to generate nice random bits is not the right one -- we really need to ensure this at runtime.

This may result in some unikernels or applications to loudly fail on your favourite platform. So then, there's hope that bugs will be reported and there are incentives to fix these.


dinosaure:
This PR seems nice and I agreed that we should check (and hard fail) when our entropy system is not good. We probably also add an option (defaults to false) to avoid this hard-failure I think. What worries me is that we're running into issues on platforms such as RISC-V and macOS. It would be good to finally find a solution for these platforms before merging this PR.

hannesm:
For me, as maintainer of the package, I would like to _not_ have an option to avoid this hard failure. It will just be used by people in their applications and leads to runtime bad behaviour (predictable random in the worst case).

I agree we should fix the platforms we're interested in supporting entropy collection. It sounds like arm64 is needed. RISC-V, and amd64 with macOS: I personally don't have any hardware, and am happy to see patches, but won't be able to dig deep into that myself.

For arm64 I'm happy to dig deeper. macOS with amd64 seems like not really worth the fix (since apple doesn't ship this platform anymore anyways).

hannesm:
So, how to move forward here? I can see the possibility to move mirage-crypto-rng-miou-unix in the similar way as the other Unix backends to use the `Mirage_crypto_rng_unix.use_default`? And then the unikernels really use the entropy test at startup...

WDYT @dinosaure? We can as well leave this for now, or have miou-unix not do the entropy check at initialization time. But after all I don't think there are drawbacks in using the Mirage_crypto_rng_unix.use_default...

dinosaure:
> I can see the possibility to move mirage-crypto-rng-miou-unix in the similar way as the other Unix backends to use the Mirage_crypto_rng_unix.use_default?

Yes, I think it's the best way, I tried to dig in on the RISC-V architecture but the rabbit hole seems bigger than expected (it seems that the extension required to fix RISC-V is not always available and optional, also I did not find yet any solution for ARM64 and Mac).

PR #283, opened 2026-07-06
mc_cycle_counter never returns the same value on ARM64 or RISCV64

dinosaure:
This patch adds a new value when we are into ARM64 and RISCV64. The goal is to never returns the same value multiple times when we call `mc_cycle_counter` which is needed to initialize Fortuna. This patch **does not** amplify the unpredictability of the value however and the values needed to initialize Fortuna are still poors but it permits to not fail.

### DIFF
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -52,7 +52,7 @@
    on these platforms.
 */
 
-#if defined (__arm__) || defined (__aarch64__) || defined (__riscv) && (64 == __riscv_xlen))
+#if defined (__arm__) || defined (__aarch64__) || (defined (__riscv) && (64 == __riscv_xlen))
 #include <stdint.h>
 static uint32_t mc_cycle_counter_calls = 0;
 #define mc_sequence() __atomic_fetch_add(&mc_cycle_counter_calls, 1, __ATOMIC_RELAXED)
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -40,7 +40,23 @@
    The executable test/test_entropy.ml tests parts of the requirements by
    calling mc_cycle_counter 10 times and comparing the output to the previous
    output.
+
+   On some platforms (arm64 cntvct_el0, riscv rdtime, the arm32 fallbacks),
+   the available timer ticks slower than the rate at which OCaml can call
+   this function, so two successive reads may observe the same timer value
+   (see https://github.com/mirage/mirage-crypto/issues/216). There, a
+   per-call sequence number is mixed in: the timer still provides the
+   entropy, while the sequence number guarantees that successive outputs
+   differ. This **does not** amplify the unpredictability of the value, but it
+   does ensure that you do not encounter any issues when initialising Fortuna
+   on these platforms.
 */
+
+#if defined (__arm__) || defined (__aarch64__) || defined (__riscv) && (64 == __riscv_xlen))
+#include <stdint.h>
+static uint32_t mc_cycle_counter_calls = 0;
+#define mc_sequence() __atomic_fetch_add(&mc_cycle_counter_calls, 1, __ATOMIC_RELAXED)
+#endif
 
 #if defined (_MSC_VER)
 #include <immintrin.h>
@@ -182,11 +198,11 @@ CAMLprim value mc_cycle_counter (value __unused(unit)) {
 #if defined (__i386__) || defined (__x86_64__) || defined (_MSC_VER)
   return Val_long (__rdtsc ());
 #elif defined (__arm__) || defined (__aarch64__)
-  return Val_long (read_virtual_count ());
+  return Val_long (read_virtual_count () + mc_sequence ());
 #elif defined(__powerpc64__) || defined(__POWERPC__)
   return Val_long (read_cycle_counter ());
 #elif defined(__riscv) && (64 == __riscv_xlen)
-  return Val_long (cycle_count ());
+  return Val_long (cycle_count () + mc_sequence ());
 #elif defined (__s390x__)
   return Val_long (getticks ());
 #elif defined(__mips__)
--- a/src/native/entropy_cpu_stubs.c
+++ b/src/native/entropy_cpu_stubs.c
@@ -52,7 +52,7 @@
    on these platforms.
 */
 
-#if defined (__arm__) || defined (__aarch64__) || (defined (__riscv) && (64 == __riscv_xlen))
+#if defined (__arm__) || defined (__aarch64__)
 #include <stdint.h>
 static uint32_t mc_cycle_counter_calls = 0;
 #define mc_sequence() __atomic_fetch_add(&mc_cycle_counter_calls, 1, __ATOMIC_RELAXED)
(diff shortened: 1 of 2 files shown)


dinosaure:
See #216

dinosaure:
At least, it's green :+1:

hannesm:
Thanks for your PR. I'm still hesitant with this, and it'll need some reading whether this is a good idea or not. Or some cryptographic expert jumps in (or someone else) and provides some references why this is a good/bad idea.

dinosaure:
Yes, unfortunately that isn’t a proper solution, and I imagine it might make the next number generated by Fortuna more predictable, given the limited data used to initialise the engine.

hannesm:
I thought a bit more while on a train about this. Indeed it makes the tests pass, this is good.

But it doesn't add any entropy.. well that's a strong statement and actually depends on the threat model. If the attacker is able to observe how often the cycle_count is called, they can track the counter and thus immediately know it's value. So it won't give anything.

On the other hand if the attacker doesn't know bow often cycle_count is called, they can't predict the counter and it is fine.

Below the line the question is about the threat model. Esp. for the architectures mentioned here, I'd guess that anyone who wants to put this rng into production would investigate some time to figure out whether some CPU randomness (such as rdseed/rdrand) is available. I suspect on arm64 at least there are instructions available that may be supported by some chips and not by others.

I don't have a good feeling about this patch-which-makes-the-test-good-but-unclear-whether-theres-any-entropy-gained. If there's more interest we can remove the test (and have #282 not merged right now (or only on architectures we know are good)). There's as well the possibility that someone provides funding and sample hardware to fix those architectures.

The other question I have is what your motivation for riscv & arm64 is? Unix miou applications? Unikernels? If the former, shouldn't mirage-crypto-rng-miou-unix use getrandom (or /dev/urandom) as the other backends since #256? (of course pfortuna can still be used, but the feed can use getrandom/dev-urandom)

dinosaure:
Ok, so the real good path will be to use `getrandom` for `mirage-crypto-rng-miou-unix` :+1:

dinosaure:
I just found an extension on RISC-V which permits  to us to have an unpredictable value to initialize our Fortuna engine (without superuser privileges). It mostly seems that `ocaml-ci` does not virtualize RISC-V 64 with such extension so I made a pull-request for that, see: https://github.com/ocurrent/obuilder/pull/217

You can see details of this extension here: https://docs.riscv.org/reference/isa/extensions/crypto-scalar/_attachments/riscv-crypto-spec-scalar.pdf (Chapter 4, The Entropy source)

I will also try to figure out for ARM64 (Mac), I'm sure we can have something better.

hannesm:
Thanks for your investigation. I'd place that in the mc_cpu_rdseed/mc_cpu_rdrand (actually, these functions would need some refactoring) -- so the architecture-specific CPU entropy -- not the cycle counter.

It is a bit messy in the code, and as well the entropy harvesting code is slightly messy, but there should be a separation:
- cycle counter (maybe not available on RISC-V? arm64?)
- CPU entropy (RDSEED/RDRAND on x86, and what you discovered here on RISC-V)

The mixing and entropy harvesting combines both cycle counter and CPU entropy in a later step...

dinosaure:
Yes, I mostly experiment and see the CI, don't think that what I propose is my last version, I can set this PR as a draft and come back when I think everything is ready.

nickbetteridge:
I've been working on my own ocaml freestanding. I hit this from the other direction (freestanding aarch64/riscv64 targets) and ended up measuring it, so here is what I found — it may save some time.

The counter is not a cycle counter on either arch

That is the whole reason mc_cycle_counter repeats. On x86 rdtsc runs at core frequency; the two branches below do not:

|arch      |current instruction|what it is                                                |resolution                                                             |
|---           |      :----:                    |  :----:                                                       |  :----:                                                                      |
|aarch64|mrs xN, cntvct_el0|the system counter, fixed frequency|typically 24–54 MHz (CNTFRQ_EL0 reports it)|
|riscv64  | rdtime                    |the time CSR, fixed frequency           | often 1–10 MHz                                                  |


At ~3 GHz, a 24 MHz counter ticks once per ~125 cycles, so a tight loop reads the same value repeatedly — exactly the symptom. Adding a increment removes the repetition but, as you say above, adds nothing to sample; and since 2.1.0 the entropy test is skipped on arm64 and and riscv, so there is currently no signal either way.

The real cycle counters exist, but are privileged

  - aarch64: PMCCNTR_EL0 counts CPU cycles. It needs the PMU enabled (PMCR_EL0.E, PMCNTENSET_EL0[31], and PMCR_EL0.D clear so it does not divide by 64) and EL0 access needs PMUSERENR_EL0.{EN,CR}. Under Linux at EL0 that is normally off — perf owns the PMU — but a freestanding guest at EL1 (solo5, xen, mkernel) can simply enable it for itself, and those are the targets where this bites. Worth having CNTFRQ_EL0 as the decision input: if the system counter is fast enough, keep using it.
  - riscv64: rdcycle is the CPU cycle counter, gated by [m|s]counteren.CY; rdtime is gated by .TM. Linux ≥ 6.6 made rdcycle privileged for userspace, which is why the hosted path uses rdtime.

Either way this needs a fallback path, since both can be unavailable.

One concrete bug: the riscv64 selector does not do what it looks like

```
  __asm__ __volatile__ ("rdcycle %0" : "=r" (rval));
#else
  __asm__ __volatile__ ("rdtime %0" : "=r" (rval));
#endif
```

A freestanding toolchain that defines neither macro silently gets the slow timer. Mine is one such: my riscv64 cross toolchain reports system: none but its C compiler is a plain gcc with no __ocaml_* define, and the object contains rdtime, confirmed by objdump. Keying on a positive build flag (or on ocaml_config_var "system" = "none" in config/cfg.ml) would be more robust than on those two macros.

On the CPU entropy half — agreed on the separation

I implemented RNDR/RNDRRS (aarch64 FEAT_RNG) and the Zkr seed CSR as Entropy.cpu_rng sources, leaving mc_cycle_counter untouched, which is the split @hannesm asks for above. Two things I learned that bear on the "not consistently available" problem:

  - Detection cannot be a runtime probe on either arch. ID_AA64ISAR0_EL1 is undefined at EL0 unless the OS emulates the read, and executing RNDR to try it is equally undefined when the feature is absent — so hosted targets should ask the OS (getauxval(AT_HWCAP2) & HWCAP2_RNG, or sysctlbyname("hw.optional.arm.FEAT_RNG")), and freestanding targets at EL1 can read the ID register directly. For Zkr there is no unprivileged feature register at all: access depends on mseccfg.{s,u}seed being delegated to the mode you run in, which is a platform property, not a discoverable one. I therefore made Zkr build-time opt-in and off by default — the same conclusion you reached, just arrived at from the privilege side.
  - FEAT_RNG is ARMv8.5+, so it fixes seeding on Graviton3-class and newer parts and not on, say, a Raspberry Pi 4. It narrows the problem rather than removing it, which is another argument for fixing the counter as well.


Happy to send the aarch64/riscv64 entropy work as a PR if it would be useful — it is compile-checked for both arches (and x86 unchanged) but not yet run on aarch64 or riscv64 hardware, which should be done over the next 2-4 weeks.

The majority of the detail work was done by Claude

dinosaure:
> The majority of the detail work was done by Claude

I'm really sorry, but this sort of message is rather incomprehensible and very difficult to read (not to mention the language barrier). We're not against the use of AI, but a proofread and a re-evaluation of what the AI is saying would be necessary and appreciated next time.

> Happy to send the aarch64/riscv64 entropy work as a PR if it would be useful

At this stage, we can indeed try to improve `mc_cpu_rdseed`/`mc_cpu_rdrand` based on what has been discovered (and avoid the whirlwind bootstrap, which uses the `cycle_counter`).

PR #285, opened 2026-07-27
Re-add blinding for NIST elliptic curves (Dsa.sign)

hannesm:
Reverts "mirage-crypto-ec: remove blinding for sign"

This reverts commit 2cdea169560c4b4ff3351a087d93dbc028d7195b.

My measured performance impact is 10-15%. Is this reasonable?

//cc @edwintorok (of course others if they have opinions)

### DIFF
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -724,7 +724,7 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
   let sign ?(mask = `Yes) ~key ?k msg =
     (* blinding: literature: s = k^-1 * (m + r * priv_key) mod n
        we blind, similar to OpenSSL (https://github.com/openssl/openssl/commit/a3e9d5aa980f238805970f420adf5e903d35bf09):
-       s = k^-1 * blind^-1 (blind * m + blind * r * priv_key) mod n
+       s = k^-1 * blind^-1 * (blind * m + blind * r * priv_key) mod n
     *)
     let b = blind mask in
     let msg = padded msg in
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -698,23 +698,10 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
     let q = S.scalar_mult_base d in
     (d, q)
 
-  let not_zero =
-    let zero = String.make Param.byte_length '\000' in
-    fun n -> not (String.equal zero n)
-
-  let mod_n v =
-    let v' = F.from_be_octets v in
-    let v' = F.mul v' F.one in
-    let v' = F.from_montgomery v' in
-    F.to_be_octets v'
-
-  let smaller_n v =
-    String.equal v (mod_n v)
-
   let blind mask =
     let rec rng g =
       let r = Mirage_crypto_rng.generate ?g Param.byte_length in
-      if not_zero r && smaller_n r then begin
+      if S.is_in_range r then begin
         let ba = F.from_be_octets r in
         Some (ba, F.inv ba)
       end else
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -55,7 +55,8 @@ module type Dsa = sig
   val pub_to_octets : ?compress:bool -> pub -> string
   val pub_of_priv : priv -> pub
   val generate : ?g:Mirage_crypto_rng.g -> unit -> priv * pub
-  val sign : key:priv -> ?k:string -> string -> string * string
+  val sign : ?mask:[ `No | `Yes | `Yes_with of Mirage_crypto_rng.g ] ->
+    key:priv -> ?k:string -> string -> string * string
   val verify : key:pub -> string * string -> string -> bool
   module K_gen (H : Digestif.S) : sig
     val generate : key:priv -> string -> string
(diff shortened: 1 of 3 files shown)


ec/mirage_crypto_ec.ml
```diff
@@ -697,6 +698,33 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
...
+    let v' = F.from_montgomery v' in
+    F.to_be_octets v'
+
+  let smaller_n v =
+    String.equal v (mod_n v)
+
+  let blind mask =
+    let rec rng g =
+      let r = Mirage_crypto_rng.generate ?g Param.byte_length in
+      if not_zero r && smaller_n r then begin
```
edwintorok: Can this be simplified to `S.is_in_range r` as done in 9d6c0b9eae077a9855d2f76db15224d87ce942cd for a similar check?

ec/mirage_crypto_ec.ml
```diff
@@ -697,6 +698,33 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
...
+        let ba = F.from_be_octets r in
+        Some (ba, F.inv ba)
```
edwintorok: The inverse calculation here seems to be the expensive part.
Instead of `k^-1 * blind^-1` perhaps we could do `(k * blind) ^ -1`?

The performance is closer to the original then.
Although I'll have to think more about whether this would defeat the purpose of blinding (at first glance it seems fine, we still introduce a random value that removes direct correlation between secrets and arithmetic operations)
 
```diff
diff --git a/ec/mirage_crypto_ec.ml b/ec/mirage_crypto_ec.ml
index eb5beacc6d..c01132618c 100644
--- a/ec/mirage_crypto_ec.ml
+++ b/ec/mirage_crypto_ec.ml
@@ -698,25 +698,12 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
     let q = S.scalar_mult_base d in
     (d, q)
 
-  let not_zero =
-    let zero = String.make Param.byte_length '\000' in
-    fun n -> not (String.equal zero n)
-
-  let mod_n v =
-    let v' = F.from_be_octets v in
-    let v' = F.mul v' F.one in
-    let v' = F.from_montgomery v' in
-    F.to_be_octets v'
-
-  let smaller_n v =
-    String.equal v (mod_n v)
-
   let blind mask =
     let rec rng g =
       let r = Mirage_crypto_rng.generate ?g Param.byte_length in
-      if not_zero r && smaller_n r then begin
+      if S.is_in_range r then begin
         let ba = F.from_be_octets r in
-        Some (ba, F.inv ba)
+        Some ba
       end else
         rng g
     in
@@ -760,20 +747,20 @@ module Make_dsa (Param : Parameters) (F : Fn) (P : Point) (S : Scalar) (H : Dige
       | Some r ->
         let r_mon = F.from_be_octets r in
         let kmon = F.from_be_octets k' in
+        let kmon =
+          match b with None -> e | Some b -> F.mul b kmon
+        in
         let kinv = F.inv kmon in
         let dmon = F.from_be_octets (S.to_octets key) in
         let dmon =
-          match b with None -> dmon | Some (b, _) -> F.mul b dmon
+          match b with None -> dmon | Some b -> F.mul b dmon
         in
         let rd = F.mul r_mon dmon in
         let e =
-          match b with None -> e | Some (b, _) -> F.mul b e
+          match b with None -> e | Some b -> F.mul b e
         in
         let cmon = F.add e rd in
         let smon = F.mul kinv cmon in
-        let smon =
-          match b with None -> smon | Some (_, b') -> F.mul b' smon
-        in
         let s = F.from_montgomery smon in
         let s = F.to_be_octets s in
         if S.not_zero s && S.not_zero r then
```

edwintorok (approved):
Thanks, the revert looks correct, some comments about improving performance (as a separate commit or PR on top of this one perhaps).

ec/mirage_crypto_ec.ml
hannesm: done in 24dc49f

ec/mirage_crypto_ec.ml
hannesm: The `s_in_range` I applied, thanks.

For blinding, I took OpenSSL approach (https://github.com/openssl/openssl/commit/a3e9d5aa980f238805970f420adf5e903d35bf09):

Instead of doing s = k^-1 * (m + r * priv_key) mod n, it does s = k^-1 * blind^-1 * (blind * m + blind * r * priv_key) mod n

But your approach, s = (k * blind) ^-1 * (blind * m + blind * r * priv_key) mod n, looks legitimate as well. Would be interesting to find literature on that; and/or other implementations doing this.

hannesm: So, botan https://botan.randombit.net/handbook/side_channels.html mentions for ECDSA: the inversion and portions of the scalar arithmetic use blinding. The inverse of k is computed as (k*z)^-1 * z, and the computation of s, normally ((x * r) + m)/k, is computed instead as ((((x * z) * r) + (m * z)) / k) / z, for a random z.

Which sounds to me that using your approach, `(k * blind) ^ -1` - is legitimate.

hannesm: done in 740db5c

edwintorok: Thanks for finding it, the comment in their source code further confirms that indeed they use the same mechanism.

Interestingly they also have a timing test: https://github.com/randombit/botan/blob/417465db10f418bdc945ed4a80d03b9d44bbf0d0/src/cli/timing_tests.cpp, which measures the computation and the results are analysed by https://github.com/seecurity/mona-timing-report.git. That is a nice idea to find (unintentional) timing leaks.

hannesm: That sounds interesting. I'm open for PRs on that front. :)

PR #286, opened 2026-07-27
Remove mirage-crypto-rng-miou-unix, so the rng can be initialized with Mirage_crypto_rng_unix.use_default ()

hannesm:
//cc @dinosaure this is a breaking change, the pfortuna is kept for the mirage-crypto-rng-mkernel implementation.

does this make sense? all users should call `Mirage_crypto_rng_unix.use_default ()` instead of `Mirage_crypto_rng_miou_unix.initialize`...


dinosaure:
Seems good for me!

hannesm:
The windows CI is failing, but this has been the case earlier...

PR #287, opened 2026-07-27
rng: remove a wrong [@@noalloc] on the getrandom binding

samoht:
The C function can fail and raise, but the binding was marked `[@@noalloc]`, which is not right. Removing the annotation on both bindings.


hannesm:
thanks

PR #288, opened 2026-07-27
rng: avoid a spurious zero-length getentropy call

samoht:
When the requested length is a multiple of 256, the loop makes one extra `getentropy(data + len, 0)` call (zero-length), so it does nothing, just an unnecessary syscall. Using `<` instead of `<=` skips it.


hannesm:
thanks

Issue #291, opened 2026-08-06
Regression in 2.1.0->2.2.0: failure in test_symmetric_runner on armhf


glondu:
Hi,

While upgrading ocaml-mirage-crypto from 2.1.0 to 2.2.0 in Debian, a new failure appeared on armhf:
```
File "tests/dune", line 8, characters 7-28:
8 |  (name test_symmetric_runner)
           ^^^^^^^^^^^^^^^^^^^^^
(cd _build/default/tests && ./test_symmetric_runner.exe)
accel: 
..................................................................E
==============================================================================
Error: All:1:Cipher:12:AEAD-forged-tag:0.

File "/build/reproducible-path/ocaml-mirage-crypto-2.2.0/_build/default/tests/oUnit-All-sbuild#00.log", line 504, characters 1-1:
Error: All:1:Cipher:12:AEAD-forged-tag:0 (in the log).

Worker stops running: Killed by signal -22
------------------------------------------------------------------------------
Ran: 67 tests in: 1.37 seconds.
FAILED: Cases: 67 Tried: 67 Errors: 1 Failures: 0 Skip:  0 Todo: 0 Timeouts: 0.
```
Signal -22 is Bus error... maybe some unaligned access? I've digged a bit, and it seems the error is in `test_case gcm` in: https://github.com/mirage/mirage-crypto/blob/5f26ca1c9284fe9c93ab23768ff88ba345b360c5/tests/test_cipher.ml#L1031

@hannesm @samoht : I've bisected, and the first bad commit is 25e7570aec91e092b347561c23f84b6ec39e7163

hannesm:
Thanks for your report. I don't have armhf hardware, and it looks like the CI doesn't run any arm32 jobs anymore. I don't quite understand how that commit can introduce the failure. Which OCaml compiler do you use? Can you debug & propose a PR? Otherwise I'd mark mirage-crypto* as unsupported on 32 bit platforms...

glondu:
> I don't have armhf hardware

I could reproduce on my (amd64) machine, using an armhf chroot (and qemu-user-static+binfmt), that's how I bisected. Understanding fully and proposing a PR myself will need some more time.

> I don't quite understand how that commit can introduce the failure.

The failing test is introduced by the same commit. Most likely, the defect was already there before the commit, just not tested.

> Which OCaml compiler do you use?

[...]

PR #292, opened 2026-08-07
Fix an unaligned access that caused a bus error on armhf

glondu:
Bug: https://github.com/mirage/mirage-crypto/issues/291


hannesm:
thanks

PR #296, opened 2026-08-10
Don't display anything when we run our test

dinosaure:
Otherwise, we see `TESTING...` when we launch our unikernels.


reynir:
Ah! I was wondering where that message suddenly came from.

hannesm:
thanks

PR #299, opened 2026-08-19
CCM: avoid padding aligned associated data

torinnd:
CCM currently appends a zero block when the encoded AAD already ends on a block boundary. This changes the tag, for example when a 14-byte AAD follows its two-byte length prefix. This patch adds no padding in that case and includes a known-answer test.

Verify the expected tag with:

```bash
python3 -m pip install cryptography
python3 - <<'PY'
from cryptography.hazmat.primitives.ciphers.aead import AESCCM
key = bytes.fromhex("000102030405060708090a0b0c0d0e0f")
nonce = bytes.fromhex("000102030405060708090a0b0c")
aad = bytes.fromhex("000102030405060708090a0b0c0d")
print(AESCCM(key, tag_length=16).encrypt(nonce, b"", aad).hex())
PY
```

Expected: `4c8deb839f1db3856356fe0c0956db30`

### DIFF
--- a/src/ccm.ml
+++ b/src/ccm.ml
@@ -41,7 +41,7 @@ let gen_adata a =
   in
   let to_pad =
     let leftover = (llen + String.length a) mod block_size in
-    block_size - leftover
+    if leftover = 0 then 0 else block_size - leftover
   in
   llen + String.length a + to_pad,
   fun buf off ->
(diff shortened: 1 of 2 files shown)


hannesm:
looks great, thanks for finding and reporting this issue.

PR #300, opened 2026-08-23
mirage-crypto-rng-mkernel: when the RNG is initialized, seed Stdlib.Random

hannesm:
cc @dinosaure @reynir


dinosaure:
I'm ok with this PR, we probably can also provide a `rng Mkernel.arg` to facilitate the usage of `mirage-crypto-rng-mkernel`

dinosaure (approved):

reynir (approved):

hannesm:
I rerun the CI, and it looks fine now. There's still some failure which is not related to this PR, though.

PR #301, opened 2026-08-25
RSA.OAEP: reject encodings without a separator

torinnd:
_(Note: I'm not a cryptographer and this was LLM-assisted. Please check my work!)_

RSA-OAEP encodes a message as:

```text
H(label) || zero padding || 0x01 || message
```

During decryption, the first nonzero byte after `H(label)` must therefore be the `0x01` separator. An encoding without that separator is invalid.

When the search found no nonzero byte, the decoder used to fall back to offset zero. If `H(label).[0]` happened to equal `0x01`, a separator-free encoding passed validation.

I think this should fall back to `hlen` instead. A failed search guarantees that this byte is zero, so the separator check fails, and this should continue to be a constant-time search.

The regression test uses a label whose SHA-1 digest starts with `0x01` and a deterministic OAEP block without a separator.


hannesm:
I checked this with the specification. Thanks a lot!

PR #302, opened 2026-08-25
mirage-crypto-rng.unix: avoid duplicate output after fork

torinnd:
`Urandom` currently reads through an `In_channel`, which retains up to 64 KiB of random data in userspace. If the process forks after a read, parent and child inherit identical unread bytes and can return the same random output.

I think it's safer to read directly from the `/dev/urandom` descriptor instead. This preserves thread safety without copying random data across forks.

The regression test primes the generator, forks, and compares output from both processes.

### DIFF
--- a/rng/unix/dune
+++ b/rng/unix/dune
@@ -17,7 +17,7 @@
  (name mirage_crypto_rng_unix)
  (public_name mirage-crypto-rng.unix)
  (modules mirage_crypto_rng_unix urandom getentropy)
- (libraries mirage-crypto-rng unix logs threads.posix)
+ (libraries mirage-crypto-rng unix logs)
  (foreign_stubs
   (language c)
   (include_dirs ../../src/native)
--- a/rng/unix/urandom.ml
+++ b/rng/unix/urandom.ml
@@ -1,4 +1,3 @@
-
 type g = Unix.file_descr
 
 let block = 1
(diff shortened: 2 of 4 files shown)


hannesm:
Thanks, I pushed a commit to your branch simplifying the implementation.

The original reasoning for using a buffer and a Mutex was performance, so here are some numbers (using OCaml 5.5.0 on FreeBSD 15.1 on a rather old Intel CPU):

Before this PR:
```
* [urandom]
       16:  99.859030 MB/s  (12986886 iters in 1.984 s)
       64:  216.250905 MB/s  (7048408 iters in 1.989 s)
      256:  305.912223 MB/s  (2521638 iters in 2.012 s)
     1024:  336.212150 MB/s  (701535 iters in 2.038 s)
     8192:  315.213262 MB/s  (83646 iters in 2.073 s)
```

After this PR:
```
* [urandom]
       16:  14.141226 MB/s  (1794702 iters in 1.937 s)
       64:  55.376279 MB/s  (1857734 iters in 2.048 s)
      256:  149.685339 MB/s  (1225722 iters in 1.999 s)
     1024:  217.583420 MB/s  (466428 iters in 2.093 s)
     8192:  282.790942 MB/s  (75691 iters in 2.091 s)
```

In contrast, the `getentropy`:
```
       16:  19.266151 MB/s  (2536114 iters in 2.009 s)
       64:  76.924333 MB/s  (2485874 iters in 1.972 s)
      256:  186.308350 MB/s  (1467487 iters in 1.923 s)
     1024:  191.817160 MB/s  (395308 iters in 2.013 s)
     8192:  186.889090 MB/s  (49576 iters in 2.072 s)
```

I see the issue with the duplicated output after fork, so I'm happy to merge this. By default, the `getentropy` provider is used anyways.

torinnd:
Looks good to me; thanks for cleaning the PR up!

PR #303, opened 2026-08-25
rng: reuse registered entropy sources

torinnd:
Entropy samples carry a numeric source ID so Fortuna can distinguish their origins.

The source set compares entries by name. Registering an existing name therefore left the set unchanged but returned a new ID equal to the set’s size. The next distinct source received the same ID.

Return the existing source for duplicate registrations. New registrations continue to use the atomic compare-and-set loop.


hannesm:
thanks

PR #304, opened 2026-08-25
CCM: reject messages that do not fit in the length field

torinnd:
CCM sets $L = 15 - \lvert nonce\rvert$ and encodes the message length in $L$ bytes. Larger lengths are currently truncated. With a 13-byte nonce, for example, 65,536 bytes is encoded as length zero.

This change validates the nonce and message length in the shared CCM core before processing any data. This should exhaustively cover all safe and unsafe encrypt/decrypt entry points.

The regression test exercises a 13-byte nonce and a 65,536-byte message.

### DIFF
--- a/src/ccm.ml
+++ b/src/ccm.ml
@@ -2,6 +2,17 @@ open Uncommon
 
 let block_size = 16
 
+let valid_nonce nonce =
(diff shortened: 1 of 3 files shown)


hannesm:
thanks

PR #305, opened 2026-08-25
Reject wrapping stream cipher counters

torinnd:
Counter-based ciphers encrypt successive counter values to produce a keystream. If the counter wraps to zero, they can produce keystream that was already used under the same key.

Check the complete request before writing output. AES and DES CTR now reject requests that exhaust their 128-bit or 64-bit counters. ChaCha20 does the same for its 32-bit and 64-bit variants.

AES-GCM is slightly different: its 32-bit increment is defined to wrap. It therefore retains that behavior but rejects requests exceeding the specified 4,294,967,294-block limit.

The tests confirm that the final available block succeeds and the next block is rejected.

### DIFF
--- a/src/chacha20.ml
+++ b/src/chacha20.ml
@@ -11,7 +11,7 @@ let of_secret a = a
 let chacha20_block state idx key_stream =
   Native.Chacha.round 10 state key_stream idx
 
-let init ctr ~key ~nonce =
+let init ctr ~key ~nonce ~blocks =
   let ctr_off = 48 in
   let set_ctr32 b v = Bytes.set_int32_le b ctr_off v
   and set_ctr64 b v = Bytes.set_int64_le b ctr_off v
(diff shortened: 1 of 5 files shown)


hannesm:
looks good. a test for GCM wrap would be nice -- but unfortunatly my laptop has not enough memory for it.

hannesm:
I do appreciate the quick PR on this. But I'm curious whether this is the right strategy.

Surely, for ChaCha20 the ctr = 0 is special since that is used as nonce. So, there I have not seen usages with a custom ctr.

For GCM here it is already "count the blocks" instead of "prevent the ctr value from wrapping".

Now, for CTR (unsure about CCM), at least OpenVPN uses the first 16 bytes of the hmac as initial counter value for each packet --> my intuition is that here a wrap is fine as long as there's no reuse of the same counter. This is actually a bit tricky to preserve in OpenVPN.

So, did you look into specific protocols or where did you look for the guards deployed here (in terms of counter wraps vs checking the number of blocks)? @torinnd

torinnd:
I think you're right and the generic CTR strategy is too restrictive. I'll cut another PR with fixes, as this one is already closed.

My LLM-assisted tentative plan is to:

- Restore modular add_ctr and counter generation for generic AES/DES CTR, including the original rollover tests and API documentation.
- Keep GCM’s block-count guard. inc32 remains modulo as required by the standard.
- Remove numeric-wrap rejection from raw ChaCha20 with caller-selected counters.
- Enforce ChaCha20-Poly1305’s mode-level limit instead: counter zero derives the Poly1305 key, data starts at one, and data must not cycle back to zero.

PR #306, opened 2026-08-26
Use block limits for stream ciphers

torinnd:
PR #305 rejected numeric counter rollover, which was a bit too conservative. Raw CTR may start at a protocol-derived value, so crossing zero is valid provided no counter repeats.

We should restore modular AES/DES CTR and raw ChaCha20 behavior, and continue enforcing the mode-level limits: $2^{32}-2$ blocks for GCM, $2^{32}$ blocks for raw IETF ChaCha20, and $2^{32}-1$ blocks for ChaCha20-Poly1305.

### DIFF
--- a/src/chacha20.ml
+++ b/src/chacha20.ml
@@ -20,12 +20,13 @@ let init ctr ~key ~nonce ~blocks =
   and inc64 b = set_ctr64 b (Int64.add (Bytes.get_int64_le b ctr_off) 1L)
   in
   let check_blocks max =
-    if Int64.of_int blocks > max then invalid_arg "Chacha20: too many blocks"
+    if Int64.unsigned_compare (Int64.of_int blocks) max > 0 then
+      invalid_arg "Chacha20: too many blocks"
   in
   let s, key, init_ctr, nonce_off, inc =
     match String.length key, String.length nonce, Int64.shift_right ctr 32 = 0L with
     | 32, 12, true ->
-      check_blocks 0xfffffffeL;
+      check_blocks 0x100000000L;
       let ctr = Int64.to_int32 ctr in
       "expand 32-byte k", key, (fun b -> set_ctr32 b ctr), 52, inc32
     | 32, 12, false ->
@@ -106,7 +107,18 @@ let mac_into ~key ~adata src ~src_off len dst ~dst_off =
                            len_buf, 0, String.length len_buf ]
     dst ~dst_off
 
+let check_aead_blocks nonce len =
+  let blocks = len // block in
+  match String.length nonce with
+  | 12 ->
+    if Int64.of_int blocks > 0xffffffffL then invalid_arg "Chacha20: too many blocks"
+  | 8 ->
+    if Int64.unsigned_compare (Int64.of_int blocks) Int64.minus_one > 0 then
+      invalid_arg "Chacha20: too many blocks"
+  | _ -> ()
+
 let unsafe_authenticate_encrypt_into ~key ~nonce ?(adata = "") src ~src_off dst ~dst_off ~tag_off len =
+  check_aead_blocks nonce len;
   let poly1305_key = generate_poly1305_key ~key ~nonce in
   crypt_into ~key ~nonce ~ctr:1L src ~src_off dst ~dst_off len;
   mac_into ~key:poly1305_key ~adata (Bytes.unsafe_to_string dst) ~src_off:dst_off len dst ~dst_off:tag_off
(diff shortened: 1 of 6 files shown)


torinnd:
@hannesm I took a quick look at this, but I expect you're probably better-equipped than I to opine about the correctness of this change. I think it looks right, but then again I thought the last one was right too!

src/chacha20.ml
```diff
@@ -19,24 +19,20 @@ let init ctr ~key ~nonce ~blocks =
...
-       && Int64.unsigned_compare (Int64.of_int (blocks - 1)) (Int64.sub max ctr) > 0
-    then invalid_arg "Chacha20: counter wrapped"
+  let check_blocks max =
+    if Int64.of_int blocks > max then invalid_arg "Chacha20: too many blocks"
   in
   let s, key, init_ctr, nonce_off, inc =
     match String.length key, String.length nonce, Int64.shift_right ctr 32 = 0L with
     | 32, 12, true ->
-      check_counter 0xffffffffL;
+      check_blocks 0x100000000L;
```
hannesm: this is the code path for below, no? so where the bound is 0xffffffff..

src/chacha20.ml
```diff
@@ -19,24 +19,20 @@ let init ctr ~key ~nonce ~blocks =
...
       let ctr = Int64.to_int32 ctr in
       "expand 32-byte k", key, (fun b -> set_ctr32 b ctr), 52, inc32
     | 32, 12, false ->
       invalid_arg "Counter too big for IETF mode (32 bit counter)"
     | 32, 8, _ ->
-      check_counter Int64.minus_one;
```
hannesm, line 36: this code path has a limit of 2^64 - 1 blocks? but that is not checked...

src/chacha20.ml
```diff
@@ -19,24 +19,20 @@ let init ctr ~key ~nonce ~blocks =
...
       "expand 32-byte k", key, (fun b -> set_ctr64 b ctr), 56, inc64
     | 16, 8, _ ->
-      check_counter Int64.minus_one;
```
hannesm, line 39: this as well uses 64 bit counter, so the number of blocks should be at most 2 ^ 64 - 1

src/chacha20.ml
```diff
@@ -108,7 +104,12 @@ let mac_into ~key ~adata src ~src_off len dst ~dst_off =
                            len_buf, 0, String.length len_buf ]
     dst ~dst_off
 
+let check_aead_blocks nonce len =
+  if String.length nonce = 12 && Int64.of_int (len // block) > 0xffffffffL then
+    invalid_arg "Chacha20: too many blocks"
```
hannesm: so why exactly do we check this here now? why not in "init" - where we check the remaining block counts as well?

hannesm:
This looks pretty fine.. but what about CTR? There's the potential danger that their counters overflow, no?

src/chacha20.ml
hannesm: see 62c0980099210167fe678f541f096c3df633abae

src/chacha20.ml
torinnd: _(with my usual "I'm not a cryptographer" caveat)_

I think init is shared with raw crypt, which permits 2^32 blocks. AEAD reserves counter zero and permits only 2^32 - 1. The separate guard applies that stricter limit and also checks decrypt before authentication, where init is not yet called.

src/chacha20.ml
hannesm: see 62c0980099210167fe678f541f096c3df633abae

src/chacha20.ml
hannesm: see 62c0980099210167fe678f541f096c3df633abae

src/chacha20.ml
hannesm: see 62c0980099210167fe678f541f096c3df633abae

hannesm:
I pushed the chacha changes to your branch. I think that should be fine now. but CTR mode is missing. So eventually we need to embed a check in the Counter structure.

but I've to go off now and will only come back late today...

torinnd:
I think the chacha change you've pushed may have some implementation flaws. My LLM assistant flagged these two findings, which I think are correct.

1. It compares a signed value against `Int64.minus_one`. Every 8-byte ChaCha call, including empty input, now raises.
2. It caps the 12-byte path at `0xfffffffe`, although the comment says `0xffffffff`. More fundamentally, raw ChaCha and AEAD need separate limits: $2^{32}$ and $2^{32}-1$ blocks respectively.

src/chacha20.ml
torinnd: Maybe this would be clearer if it was a named constant?

let raw_ietf_max_blocks = 0x100000000L
let aead_ietf_max_blocks = 0xffffffffL

or something like that?

torinnd:
I've landed some updates, including the total block check stuff you'd requested. Let me know what you think!

hannesm:
I think this looks good, thanks @torinnd

PR #307, opened 2026-09-10
move Wycheproof tests to jsont

hannesm:
no more yojson, and ppx_deriving


hannesm:
The reasoning behind this PR is structure & compilation time. Another goal is to not rely on ppx_deriving in any project that I'm maintaining. Any positive or negative opinion can be raised here.

dinosaure (approved):

PR #309, opened 2026-09-12
fix(uncommon): make invalid_arg domain safe

thevilledev:
## Summary

`Uncommon.invalid_arg`  prints to [`Format.str_formatter`](https://ocaml.org/manual/5.5/api/Stdlib.Format.html#VALstr_formatter) but reads the result back through [`Format.flush_str_formatter`](https://ocaml.org/manual/5.5/api/Stdlib.Format.html#VALflush_str_formatter).

On OCaml 5 the first is the initial domain's formatter and the second returns the current domain's buffer. Raised on a spawned domain, the exception carries an empty message. The text stays in the shared formatter, and the next rejection on the initial domain reports both. The leftover grows until then, and concurrent domains write to the formatter unsynchronised.

The function is on the error path for every rejected key, nonce and tag in AES-GCM and ChaCha20-Poly1305.

Discovered by running `mirage-crypto` in parallel across domains.

## Changes

[`Format.kasprintf`](https://ocaml.org/manual/5.5/api/Stdlib.Format.html#VALkasprintf) allocates a buffer per call, so use it instead. The public API and the messages on the initial domain are unchanged.

The test is committed before the fix and reproduces the bug on its own. Since OCaml >= 5.0.0 is required I added a run rule and an `enabled_if` stanza to stop the test from running and compiling on older OCaml environments.

## Tests

Reject an empty AES-GCM nonce on a spawned domain, then on the main domain, and expect the same message from both. Built only on OCaml >= 5.0.0.

At the first commit `test_domains` fails:

```
expected: (spawned: "Mirage_crypto: GCM: invalid nonce of length 0", main: "Mirage_crypto: GCM: invalid nonce of length 0")
but got: (spawned: "", main: "Mirage_crypto: GCM: invalid nonce of length 0Mirage_crypto: GCM: invalid nonce of length 0")
```

At the second commit all tests pass.


hannesm:
this looks plausible, thanks. I guess that code predates kasprintf existence (since 4.03 if i see it right).

thevilledev:
Yeah, and pre-dates the OCaml domains altogether. Original commit is from 2017: https://github.com/mirage/mirage-crypto/commit/70a31c2d9f15bd04d7ccdb75223f92c4c51416ac#diff-201fb52857ff827b9df164c79f18aced192e7b3f2379c620b7abe3abd12a0220

PR #311, opened 2026-09-21
remove ARC4

hannesm:
there's been no usage, the code has potential timing leaks


avsm (approved):
I agree that remove of ARC4 is the safest option, and has precedent in other crypto libraries: https://github.com/transmission/arc4/blob/master/README.md

